PR #20, opened 2015-12-29
Update Tile.php

astersx:
This will return false if getBlock() is null, so the server wont crash!!! 
So if you think that will fix the crash if by any chance getBlock() is null,  im happy i helped....
If im wrong please no insults 😝.

### DIFF
--- a/src/pocketmine/tile/Tile.php
+++ b/src/pocketmine/tile/Tile.php
@@ -148,7 +148,8 @@ abstract class Tile extends Position{
 	 * @return \pocketmine\block\Block
 	 */
 	public function getBlock(){
-		return $this->level->getBlock($this);
+if($this->level->getBlock($this) == null){ return false;} else{
+		return $this->level->getBlock($this);}
 	}
 
 	public function onUpdate(){


astersx:
This is line 151 wich is false. return $this->level->getBlock($this);
if i check $this->lever->getBlock($this) is null?
if yes all function to be false if yes continue normal code.

im not soo good in php but i created a online php game in 1 week just with if and else :).
And database was similar with pocketmine (players folder). 
Anyway if this code fail,  i will try in another way to fix this.(im already using it on my sv)

PeratX:
@astersx Does your server has Trapped Chests?

astersx:
I put a trapped chest down from creative, changed to survival break it and took it!!!  Is like normal chest, so why u ask me.

PeratX:
Because I think maybe something wrong with the Trapped Chest...

astersx:
Ya,  but i edit pocketmine i use with code i comitted,  sooo :D.
Anyway until now no crashes so im good.

PeratX:
@astersx We will test it later. Thx for your report!

PR #40, opened 2016-01-01
Dog194 mul language rus

dog194:
Add genisys_rus.yml(wait for translation).
Ensure "genisys.yml" will create.


dog194:
@vellrya You can make some translation work for genisys.yml in genisys_rus.yml.

PR #133, opened 2016-01-23
Addaed ores to the Nether generator

mixailenko5:


Muqsit:
Soulsand...?

mixailenko5:
@Muqsit Soulsand and Gravel its not ores. It is cover. And i add them in future.

PR #139, opened 2016-01-24
Add a little translate and disable the PocketMine auto-updating.

GalvinGao:
@PeratX I'm not sure that hat i am edit is right or wrong so i created a pull.


CrayonDing:
Thx.

PR #144, opened 2016-01-24
php7 updates :)

xpyctum:


xpyctum:
@PeratX :+1:

PeratX:
@xpyctum And more methods can have PHP7 update:) Thx!

PR #152, opened 2016-01-24
Fix Fatal error: Declaration of pocketmine\block\PackedIce::getName()…

xpyctum:
… must be compatible with pocketmine\block\Block::getName()


flaxues:
The error is gone now!

PR #239, opened 2016-02-05
Russian translation! @rvachvg

rvachvg:
It is a Russian translation of core messages provided by @rvachvg. Useful because there is many Russian servers.


gaoyichuan:
Nice work!

PR #250, opened 2016-02-09
biome improve

dog194:


mixailenko5:
@dog194, do you know Japan language?

dog194:
@mixailenko5  I dont know Japanese. so i use English temporary. and wait someone to translate it.

PR #255, opened 2016-02-09
Change startup script slightly, add more command aliases and change s…

RandomAltThingy:
…ome messages


RandomAltThingy:
**I changed startup script so you can have a file named "Genisys.phar" instead of having to be like "Genisys_1.0.phar". Looks like I forgot to commit the bat file.**

PR #262, opened 2016-02-10
Some changes

RandomAltThingy:


gaoyichuan:
Well, we have a compile.sh just made for Genisys.. @PeratX put it here

PR #269, opened 2016-02-12
Php5 support for 0.13.0 , 013.2 and 0.14.0 build 1-6

Driesboy:


src/pocketmine/Player.php
```diff
@@ -2309,7 +2309,7 @@ public function handleDataPacket(DataPacket $packet){
 					break;
 				}
 
-				if($packet->protocol1 !== ProtocolInfo::CURRENT_PROTOCOL){
+				if(!in_array($packet->protocol1, ProtocolInfo::ACCEPTED_PROTOCOLS)){
 					if($packet->protocol1 < ProtocolInfo::CURRENT_PROTOCOL){
```
PEMapModder: Maybe you should change this line to `if($packet->protocol1 < min(ProtocolInfo::SUPPORTED_PROTOCOLS))` and similar for the other end.

Driesboy: Ok

Muqsit: Wait.... 0.13.x and 0.14 players can join a particular server?

PR #274, opened 2016-02-12
Update Chest.php

GAalexandres:
Fix chest destroy bug if you make sand fall on it.
Thanks to @aodzip

### DIFF
--- a/src/pocketmine/block/Chest.php
+++ b/src/pocketmine/block/Chest.php
@@ -41,6 +41,10 @@ class Chest extends Transparent{
 		$this->meta = $meta;
 	}
 
+	public function isSolid(){
+		return false;
+	}
+
 	public function canBeActivated(){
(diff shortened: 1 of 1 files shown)


RandomAltThingy:
I thought a chest was solid? It's a TileEntity and not an Entity and Tiles are solid so wouldn't it be solid.

PEMapModder:
Tiles have nothing to do with solid.
See also: brewing stand, sign, daylight sensor

And Entity has no solid nor non solid.

gaoyichuan:
I don't think chests can be solid.. closing

PocketKiller:
@gaoyichuan Chest::isSolid() returned false.

GAalexandres:
Try it yourself please.

gaoyichuan:
@PocketKiller So you think we should merge this?

PR #373, opened 2016-02-21
Can you make run for people who has win 7 and php 7 don't work for us

GrimDevs:
Can you like code a update for the people who has window 7 and there pic can't be upgraded to win 10 or 8 because there are not compatible to use it if you could code it I. PHP5 it could help us instead of go out beying new ones that have the same ram and cpu I mean I have me pic it off because me server it stuck on 0.13.2 and I shan't to play I thick that should get php 7 for win 7 I am on win 7 plz reply and let other also think ok


gaoyichuan:
I can tell you that PHP7 Binary works for win7

GrimDevs:
Ok then make the a zip of it for me then k can get it because it been hell for me try to find php7

GrimDevs:
I had heel
 Try to find php 7

gaoyichuan:
@GrimDevs See this
https://github.com/iTXTech/PHP-Genisys

PR #394, opened 2016-02-22
Add recipe for crafting Iron Door.

markkrueg:
This adds ability to craft Iron Door from Iron Ingots.


gaoyichuan:
Would you like to become a collaborator?

PR #421, opened 2016-02-25
Fixed hardness and tool

Muqsit:


PeratX:
Check here https://github.com/iTXTech to join org

PR #445, opened 2016-02-27
Update BrewingStand.php

erkam2002:


src/pocketmine/block/BrewingStand.php
```diff
@@ -74,31 +73,20 @@ public function getName() : string{
 	public function onActivate(Item $item, Player $player = null){
 		if($player instanceof Player){
 			//TODO lock
-			if($player->isCreative() and $player->getServer()->limitedCreative){
-				return true;
+			if($player->isCreative(){and $player->getServer()->limitedCreative				return true;
```
flaxues: Something wrong here? It gives a crashdump!

flaxues:
You should check line 76 :P
```
[09:24:02] [Server thread/CRITICAL]: ParseError: "syntax error, unexpected 'and' (T_LOGICAL_AND)" (EXCEPTION) in "/src/pocketmine/block/BrewingStand" at line 76
```

gaoyichuan:
```php
{and $player
```
So you can see where is wrong

PR #458, opened 2016-02-28
Update the constant codename to PHP7 - Genisys.

SirGamer:


PocketKiller:
why do you ident with spaces...?

SirGamer:
That's my way. :)

Issue #616, opened 2016-03-16
Hangs on restart


flaxues:
#### Issue description

When I use /stop the server just stays on "Server has stopped"

#### Steps to reproduce the issue

Stop the server with /stop

#### OS and versions

* Genisys: Latest version available on github.com
* PHP: 7
* OS: Ubuntu 15.04

#### Crashdump, backtrace or other files

```
There is no crashdump
```

gaoyichuan:
A known issue

flaxues:
I just upgraded from a older genisys because of some problems with raklib they should be fixed now but now I have this issue which is really annoying and is going to drasticly drop my uptime

flaxues:
In other words you won't fix this.

flaxues:
This is occuring since commit https://github.com/iTXTech/Genisys/commit/02af2a5bac70ff5ee29147a2284f379ef115895b

PeratX:
@flaxues Did it fixed?

flaxues:
I will try it in about 1-2 hours when I'm home

FUGAMARU:
I suddenly rude.
But in my server it has a "DO_LOOP" of "start.sh" to "yes", the server does not start automatically.
Do you have a relationship This is a bug that has been reported here?

FUGAMARU:
By the way, the server by pressing the Enter key in a state where the server does not start automatically will start to react to it.

flaxues:
Yup same problem

FUGAMARU:
I see

gaoyichuan:
That's not fixed, confirmed

PeratX:
Same as another issue.  Close

PR #618, opened 2016-03-16
Revert "Try again to fix CommandReader"

flaxues:
Revert https://github.com/iTXTech/Genisys/commit/3bcdb2a31c450512a2284af9b6918b5b82b5fc44

Fixes issue #616


gaoyichuan:
That's not the cause

flaxues:
It is because my server doesn't hang on stopping anymore

PR #629, opened 2016-03-19
Fix bug with empty creative inventory after full reload of server

mixailenko5:


ishitatsuyuki:
Look likes a hack. Can you explain?

mixailenko5:
@ishitatsuyuki  If player in creative mode, then his items updates and he can see their. I fix this bug in pm++. And i am creating server on Genisys now. And I saw this bug

mixailenko5:
I write /save-all and "sudo reboot" (in terminal). And palyers who was in creative mode dont see their items in creative inventory. Creative becomes completely empty. And this code fix this bug.

ishitatsuyuki:
Is setting the gamemode necessary? It can cause additional glitches.

PeratX:
Have tested the code without this? I think this will not happen...

mixailenko5:
@PeratX I decided to rest from PluginMaker. And i started create the srever with my mini game. But i saw bug with creative items. After rebooting system all items disappeared. I changed gamemode to survival and to creative (/gamemode 0   /gamemode 1). And it is worked. And i write this code, who fixed this bug.

mixailenko5:
Sorry for my english, but this codefix bug with loss of items on creative inventory

PeratX:
But you can see PlayerInventory.php . We already send creative inventory!(no lose)

mixailenko5:
@PeratX i saw. But i saw bu today too. And you decide take or not take this code

PeratX:
I will try to find better solution

PR #725, opened 2016-04-02
Player::removeCreativeItem

xpyctum:
Add forgotten function C:


PeratX:
Tested? Seems not right.

gaoyichuan:
Shouldn't there be something like refresh the inventory?

PR #738, opened 2016-04-04
Update Player.php

cre2016:


ishitatsuyuki:
Not sure if this should have a chance. Maybe we could just remove the chance system.

PeratX:
no.

PR #747, opened 2016-04-06
Green "0.14.1" text in server list

AryToNeX:
It must be this way in my opinion; many players don't want to join my server because they think it is incompatible with 0.14.1


Driesboy:
now 0.14.0 is red

erkam2002:
use 0.14.1 :D

markkrueg:
I tested these changes and it seems to cause problems for 14.0 players (they cannot connect). It keeps giving the "blocked for 5 seconds" thing. 

So I recommend not accepting this Pull until further testing can be made.

gaoyichuan:
Already added by @PeratX

PR #794, opened 2016-04-13
Fixbug for FlintSteel

dog194:
when FlintSteel create the Nether Portal successfully. It should not create the fire block at the same time.
就是打火石搞地狱门不会附带一个火焰了。成功创建门就直接return


ishitatsuyuki:
Code review -1

Mismatch vanilla behavior

PR #802, opened 2016-04-13
Update Info.php

erkam2002:
0.15.0 build 1


dktapps:
This won't work, 0.15.0 uses protocol 60.

erkam2002:
And now???

PR #808, opened 2016-04-14
Fixes a CommandSelector bug

Muqsit:
When a non-op sends a command, it no longer sends the player a "You don't have permission to use Command Selector!" message.


Muqsit:
Approved!

Muqsit:
Ignore the reference :3

ishitatsuyuki:
Code review +1

PR #820, opened 2016-04-16
Extend API even more!

Muqsit:
This will allow players to code custom enchantment plugins and etc. (P.S I am coding one and this is an essential). I won't be merging tho, players have different preferences! :/


dog194:
I think you solve your problem in a wrong way. If your plugin want to add something. You should change   /src\pocketmine\item\enchantment

dog194:
This is to make sure player do not input the wrong param. and to aviod some other problem. So -1

Muqsit:
@dog194 If you know how to change it through a plugin, please let me know :3

dog194:
@Muqsit 0.0 I dont know. Maybe you need to ask others.

gaoyichuan:
Maybe we need an API for managing custom enchantments .

PR #843, opened 2016-04-22
FIx DeadBush can be placed at any block

dog194:
DeadBush can be placed only at SAND PODZOL HARDENED_CLAY and STAINED_CLAY


CrayonDing:
+1

dog194:
+1

PR #848, opened 2016-04-23
Reimplement console reader and readline

ishitatsuyuki:
Completely resolves all problems.

### DIFF
--- a/src/pocketmine/PocketMine.php
+++ b/src/pocketmine/PocketMine.php
@@ -481,8 +481,6 @@ namespace pocketmine {
 
 	echo "Server has stopped" . Terminal::$FORMAT_RESET . "\n";
 
-	@kill(getmypid());
-
 	exit(0);
 
 }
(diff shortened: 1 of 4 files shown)


Muqsit:
:o:

ishitatsuyuki:
Everything ready. I think it can be approved.

ishitatsuyuki:
@Muqsit PTAL.
By the way, this will expose more PHP bugs, so the application tends to output more crash.

CrayonDing:
+1

SleepSpace9:
For a long time I'm using an init.d script which starts the start.sh:
sudo --user "$USER" -s "start.sh" > /dev/null &

Since this commit (https://github.com/iTXTech/Genisys/pull/848)  I'm not able to start Genisys Pocketmine this way. The server immediately stops after preparing the levels.

Last Entries after stop:
...
[15:49:10] [Server thread/INFO]: Unloading level "world"
[15:49:10] [Server thread/INFO]: Stopping other threads
Server has stopped
**zend_mm_heap corrupted**

But I'm still able to start the server directly with start.sh from bash commandline (not through a further script).

Any ideas or recommendations how to run pocketmine as a service on linux?

Genisys version 1.1dev "Ikaros" (API 2.0.0)
PHP: 7.0.3
OS: Ubuntu Linux

ishitatsuyuki:
You're closing stdin. This is normally considered by stop and this change adopt this behavior.

Allocate a pty: use Docker or just something like tmux.

SleepSpace9:
Thanks for your explanation and advice, I'll try that. tmux sounds interesting, so I can get to my genisys console without using rcon.

PR #857, opened 2016-04-25
Refactor block internal API

ishitatsuyuki:
Probably fixes fire spreading

### DIFF
--- a/src/pocketmine/block/Block.php
+++ b/src/pocketmine/block/Block.php
@@ -749,15 +749,15 @@ class Block extends Position implements Metadatable{
 		return 0;
 	}
 
-	public function isBlockTopFacingSurfaceSolid(Block $block){
-		if($block->isSolid()){
+	public function isTopFacingSurfaceSolid(){
+		if($this->isSolid()){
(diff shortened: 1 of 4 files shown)


ishitatsuyuki:
Not tested: please take a look.

Muqsit:
I can't say that the code doesn't do anything because it seems like im the only one facing a no-fire-spread-feature problem :3

ishitatsuyuki:
Can you test if this doesn't crash and works as intended? Thanks for the help.

Muqsit:
It doesn't cause any crashes but fire-spread never worked for me so I can't judge :/

ishitatsuyuki:
@PeratX @CrayonDing @gaoyichuan friendly ping, please take a look.

jellyfishcoder:
@Muqsit > I can't say that the code doesn't do anything because it seems like im the only one facing a no-fire-spread-feature problem :3
You are not alone, same for me. I just wanted to see what would happen if it was turned on, but it didnt work. I wont use it that much though, cause grievers.

PR #874, opened 2016-04-27
Update to Minecraft 0.14.2

RedstoneAlmeida:


Muqsit:
The protocol is incorrect.

RedstoneAlmeida:
update to new protocol...

Muqsit:
The protocol must be 60 and even after it is 60, there are block break/interact packets that have to be changed!

PR #911, opened 2016-05-02
Update Item.php

DavidJBrockway:
Fixed Slime Block position in inventory to normal mcpe


DavidJBrockway:
button problem not my code

DavidJBrockway:
button problem not me

PR #937, opened 2016-05-07
New installer options and spell correction

XFuryMCPE:
Added more options to the installer. I don't know any other languages so I didn't translate anything. Fixed "Residence" on the fire resistance potion also


Muqsit:
Nice but potion of fire residence sounded better :3

PR #952, opened 2016-05-09
Fix bug of still liquid

ishitatsuyuki:
Still water updates on adjacent block change. This is not a perfect fix though. Some server implementations use non-still one everywhere.


ishitatsuyuki:
This hasn't be tested. @Muqsit @gaoyichuan @PeratX Please take a look.

PeratX:
This won't fix that

ishitatsuyuki:
Have you tested?

PR #955, opened 2016-05-09
Added CustomBlockData to more blocks

jojoe77777:
Cauldron, FlowerPot, ItemFrame and MonsterSpawner
These were the tiles that I think need it the most (out of the ones not added yet)


ishitatsuyuki:
Approved.

Issue #984, opened 2016-05-19
0.14.3 support


Shamil0402:
Please update

xpyctum:
Already -_-

erkam2002:
Its not updated look at newest build!

kabluinc:
it was updated but there were block placing problems when in s mode so reverted back to 0.14.2

Skyladd:
Oh no the protocol changed. Hope they find it out soon

CLiu13:
I will submit a formal PR very soon.

Muqsit:
#988

Bidou0418:
cool

NML32:
Good news

Issue #983, opened 2016-05-19
Can't place blocks in Survival


legoboy0215:
#### Issue description
Title seriously says it all.

#### Steps to reproduce the issue
Join, break a block and place it.

#### OS and versions
This server is running Genisys 1.1dev-167baef "Ikaros" implementing API 2.0.0 for Minecraft: Pocket Edition v0.14.3 alpha (protocol version 70)
* Genisys: 1.1
* PHP: 7.0.2
* OS: Ubuntu

#### Crashdump, backtrace or other files
None

jojoe77777:
I have this issue too, just happened after #981 (Whoops)

legoboy0215:
Yes I can see that. Same thing that happened last time maybe?

SuperCraftFacs:
Same, happening to me. The only server that is fun right now (that is mine) is like my creative server. A lot of people don't like creative servers. Well please fix this soon. 
Btw, Lego ur server is cool and u r a great dev <3

legoboy0215:
Please don't go off-topic. You won't want to anger the staff here XD

PR #988, opened 2016-05-19
Added Support for 0.14.3 (With Updated Protocol!) [TESTED]

CLiu13:
All credit for this solution goes to the original devs who fixed it.

I have not tested this myself yet, but I will do that as soon as I can.


CLiu13:
Tested it, and everything seems to work fine!

Muqsit:
Thank you and @orlando092 !

Muqsit:
#983 #984

Leo3418:
These edits are incompatible with 0.14.2 and below versions

PR #995, opened 2016-05-20
*Workaround* (not fix) for broken tip messages in many plugins

dktapps:
Nothing fancy, simply redirects sendTip() to sendPopup(). Not ideal, but it works.


ishitatsuyuki:
Looks not good to me. I don't like to hack for just a client bug.

Plugins should be kept up to date though.

Driesboy:
but i find fix send tip or remove send tip

PR #1010, opened 2016-05-21
Real fix sendTIP

robske110:
This fix is actually working, while not breaking compatibility for plugins


ishitatsuyuki:
Well, this is just a hack though.

ImagicalGamer:
Yes but @ishitatsuyuki it works (kinda) many plugins use Tip and Popup so they can send multiple messages, this would fix compatibility for those.

PhilipphatesPHP7:
it would work!

ishitatsuyuki:
Your neighbor PR has a more beautiful patch, but closed since it's just a hack.

PhilipphatesPHP7:
ClearSky has the same code for sendTip() and it nicely works !

ishitatsuyuki:
It doesn't matter if it works or not. This is a hack, using a deprecated thing in PE.

PhilipphatesPHP7:
ohh please :c

PhilipphatesPHP7:
Yes why do somthing that works ?
Have I change to ClearSky ?

robske110:
I do know that this is a hack, but PeratX's hack doesn't even WORK?!

ishitatsuyuki:
Okay, I will revert that...

PR #1055, opened 2016-05-26
Allow proper time continuum (time past 24000), fix monsters not burning outside of 0-24000 time

dktapps:
Currently, time is reset to 0 whenever it goes past 24,000. I have enabled time past 24000 so that lunar cycles can progress properly.
I've also fixed issues where zombies and skeletons would not burn correctly when the time went past 24000

### DIFF
--- a/src/pocketmine/entity/Zombie.php
+++ b/src/pocketmine/entity/Zombie.php
@@ -128,8 +128,12 @@ class Zombie extends Monster{
 		$hasUpdate = parent::onUpdate($currentTick);
 
 		if($this->isAlive()){
-			$time = $this->getLevel()->getTime();
-			if(($time >= 0 and $time < 14000) or $time >= 23000) $this->setOnFire(2); //僵尸起火
+			/* Don't use time directly
+			 * Instead, get remainder of current time divided by 24,000
+			 * This tells us the time of day, which is what we really need
+			 */
+			$timeOfDay = abs($level->getTime() % 24000);
+			if(0 < $timeOfDay and $timeOfDay < 13000) $this->setOnFire(2); //僵尸起火
 			
 			$p = $this->getNearestPlayer();//找到最近的可以被仇恨的玩家
 			if(!$p) {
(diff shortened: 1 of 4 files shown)


PeratX:
Why you disabled time re-cycle?

dktapps:
It's unneeded in my opinion. Locking the time to the first day is in my opinion a daft hack to fix bugs in bad code.
Enabling it to go past 24,000 allows the moon phases to progress properly as they do in normal Minecraft. Now we can have day 2, day 3, etc.

dktapps:
I've added some comments to my edits to clarify

ishitatsuyuki:
From the Minecraft Wiki:
`DayTime: The time of day. 0 is sunrise, 6000 is mid day, 12000 is sunset, 18000 is mid night, 24000 is the next day's 0. This value keeps counting past 24000 and does not reset to 0.`

I think following the vanilla behavior is better.

ishitatsuyuki:
This will be rejected until you correct it to vanilla behavior. Please also squash the commits by rebasing.

dktapps:
I've squashed the commits. I don't know what you mean about the vanilla behaviour - as fair as I know this patch _does_ return it to vanilla behaviour by allowing time to progress past 24000.

ishitatsuyuki:
Oh, that was my mistake. Approved.

dktapps:
Added another commit to fix wrong burn times (zombies burning after sundown)

Muqsit:
+1

dktapps:
```
	public function SkeletonFire() {
		foreach ($this->AIHolder->getServer()->getLevels() as $level) {
			foreach ($level->getEntities() as $zo){
				if ($zo instanceof Skeleton) {
					//var_dump($p->getLevel()->getTime());
					/* Don't use time directly
					 * Instead, get remainder of current time divided by 24,000
					 * This tells us the time of day, which is what we really need
					 */
					$timeOfDay = abs($level->getTime() % 24000);
					if(0 < $timeOfDay and $timeOfDay < 13000){
						$v3 = new Vector3($zo->getX(), $zo->getY(), $zo->getZ());
						$ok = true;
						for ($y0 = $zo->getY() + 2; $y0 <= $zo->getY()+10; $y0++) {
							$v3->y = $y0;
							if ($level->getBlock($v3)->getID() != 0) {
								$ok = false;
								break;
							}
						}
						if ($this->AIHolder->whatBlock($level,new Vector3($zo->getX(), floor($zo->getY() - 1), $zo->getZ())) == "water") $ok = false;
						if ($ok) $zo->setOnFire(2);
					}
				}
			}
		}
	}
```
To clarify:
```
foreach ($this->AIHolder->getServer()->getLevels() as $level) {
```

PR #1057, opened 2016-05-26
fix name for non-block items created by name

udwarf:
Correcting bug  that If the item was created by name, than the name was set as 'Unknown'


ishitatsuyuki:
Looks good to me.

PeratX:
+1

PR #1074, opened 2016-05-30
+GetLogger Replacing SenderMSG

dc4code:
+GetLogger Replacing SenderMSG


robske110:
Why? The command sender should become the message

gaoyichuan:
No, that should be announced as a message, not as log

robske110:
even if this PR results into a crash

PR #1088, opened 2016-05-31
Fixed death when landing on Slime Blocks

dktapps:
This commit fixes players being killed by fall damage when bouncing on Slime Blocks.

~~Also removed some broken code that was causing players to fall too fast when bouncing.~~
EDIT: see below

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1926,6 +1926,8 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 						$this->inAirTicks = 0;
 					}else{
 						if(!$this->allowFlight and $this->inAirTicks > 10 and !$this->isSleeping() and $this->getDataProperty(self::DATA_NO_AI) !== 1){
+							//expectedVelocity here is not calculated correctly
+							//This causes players to fall too fast when bouncing on slime when antiFly is enabled
 							$expectedVelocity = (-$this->gravity) / $this->drag - ((-$this->gravity) / $this->drag) * exp(-$this->drag * ($this->inAirTicks - $this->startAirTicks));
 							$diff = ($this->speed->y - $expectedVelocity) ** 2;
 							if(!$this->hasEffect(Effect::JUMP) and $diff > 0.6 and $expectedVelocity < $this->speed->y and !$this->server->getAllowFlight()){
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -29,6 +29,7 @@ use pocketmine\block\Fire;
 use pocketmine\block\Portal;
 use pocketmine\block\PressurePlate;
 use pocketmine\block\Water;
+use pocketmine\block\SlimeBlock;
 use pocketmine\event\entity\EntityDamageEvent;
 use pocketmine\event\entity\EntityDespawnEvent;
 use pocketmine\event\entity\EntityLevelChangeEvent;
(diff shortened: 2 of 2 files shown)


src/pocketmine/Player.php
```diff
@@ -1924,12 +1924,9 @@ public function onUpdate($currentTick){
...
 					}else{
 						if(!$this->allowFlight and $this->inAirTicks > 10 and !$this->isSleeping() and $this->getDataProperty(self::DATA_NO_AI) !== 1){
-							$expectedVelocity = (-$this->gravity) / $this->drag - ((-$this->gravity) / $this->drag) * exp(-$this->drag * ($this->inAirTicks - $this->startAirTicks));
-							$diff = ($this->speed->y - $expectedVelocity) ** 2;
-							if(!$this->hasEffect(Effect::JUMP) and $diff > 0.6 and $expectedVelocity < $this->speed->y and !$this->server->getAllowFlight()){
-								$this->setMotion($this->temporalVector->setComponents(0, $expectedVelocity, 0));
+							if(!$this->hasEffect(Effect::JUMP) and !$this->server->getAllowFlight()){
 								/*if($this->inAirTicks < 1000){
-
+
```
ishitatsuyuki: Correct this formatting.

ishitatsuyuki:
Some formatting issues and I don't think what you are removing is problematic. -1.

dktapps:
The code I removed from Player.php was causing players to fall too fast when bouncing on slime. I'll fix the formatting

dktapps:
I've reverted the changes I made to Player.php as I realized that they broke anti-fly completely. I've added comments in instead to note that:
```
$expectedVelocity = (-$this->gravity) / $this->drag - ((-$this->gravity) / $this->drag) * exp(-$this->drag * ($this->inAirTicks - $this->startAirTicks));
```
calculates the fall velocity incorrectly, which causes issues when bouncing when antiFly is turned on.

ishitatsuyuki:
Oh? I think bouncing is making a up velocity, and it should work.

dktapps:
You'll need my patch for Entity.php to try this without getting killed, but try this
(if you want to try this without merging, I have a phar you can use to test on my [Google Drive](https://drive.google.com/open?id=0By8Xy-T6X5Z8VzNFc18ydGN5QlU) )

1. Make sure `anti-fly` is set to `false` in the `genisys.yml`.
2. Fall on slime from ~20 blocks up or so in survival. Watch how high you bounce when you fall, and how fast you fall.
3. Then turn`anti-fly` to `true` and restart the server
4. Repeat step 2. After the initial drop, you'll notice the difference - players fall too fast, and as a result bounce too high.

src/pocketmine/entity/Entity.php
```diff
@@ -1063,11 +1064,10 @@ public function fall($fallDistance){
...
 		}
 		$damage = floor($fallDistance - 3 - ($this->hasEffect(Effect::JUMP) ? $this->getEffect(Effect::JUMP)->getAmplifier() + 1 : 0));
-		foreach($this->getBlocksAround() as $block){
-			if($block->getId() === Block::SLIME_BLOCK){
-				$damage = 0;
-				break;
-			}
+		
+		//Get the block directly beneath the player's feet, check if it is a slime block
+		if($this->getLevel()->getBlock($this->floor()->subtract(0, 1, 0)) instanceof SlimeBlock){
```
ishitatsuyuki: I prefer getSide.

ishitatsuyuki:
I think you should add a up velocity so the server will calculate it correctly.

dktapps:
@ishitatsuyuki I can't use getSide here, that code gets a block relative to the player's position as opposed to on the face of another block. And it's not the up velocity that's the problem; it's that the down velocity is mis-calculated. I'm not sure how to fix that.

ishitatsuyuki:
This is approved. The bug you mentioned may be fixed by someone else as long as the comment is there.

[Approved with PullApprove](https://pullapprove.com/iTXTech/Genisys/pull-request/1088/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=iTXTech/Genisys)

ishitatsuyuki:
Oh, wait: please use rebase instead of merge.

ishitatsuyuki:
Here's a approve again, also a friendly ping: @iTXTech/developers

dktapps:
I've rebased the pull request. That merge commit was an accident. (Still learning about Git)

gaoyichuan:
Approved

ishitatsuyuki:
+1, merge

PR #1093, opened 2016-06-03
Temporary fire breaking fix.

IrishUHC:
This is tested and it does work. Like the title says it temporarily fixes breaking fire.


ishitatsuyuki:
No, you're modifying the wrong file.

IrishUHC:
My bad, let me fix that :)

ishitatsuyuki:
No, you're not testing and even worse, you shouldn't change the event code.

IrishUHC:
I am trying to figure out what to put it under, sorry for the inconvenience. First time fixing stuff like this :P

PR #1097, opened 2016-06-03
Update PocketMine.php

erkam2002:
:)


flaxues:
This won't do anything...

gaoyichuan:
Versions should only be changed by main developers

MasterlazorX:
If you're an idiot, don't change it. There is already a branch for 0.15.

PR #1111, opened 2016-06-05
Surpass time_sleep_until error when server has lag

legoboy0215:
A lot of people experience this error if their server is lagging severely:
```
 An E_WARNING error happened: "time_sleep_until(): Sleep until to time is less than current time" in "/src/raklib/server/SessionManager" at line 109
```
I believe this surpasses the warning/error.


ishitatsuyuki:
Yeah, that's OK. Approved.

Muqsit:
+1

PR #1112, opened 2016-06-06
Fixed sendTip and sendPopup

MasterlazorX:
Tested on Genisys prior to fixing, sendTip and sendPopup didn't work,
Tested after fixing the issue, sendTip and sendPopup worked fine.


dog194:
-1 
Note for plugin developers: Do NOT use this function anymore, it's deprecated

dog194:
And tested, sendPopup works fine.

PR #1133, opened 2016-06-11
Dev. add new potion arrows

SuperWangCC:
I add some potion arrows.
But without luck arrow because the libraries 'effect' didn't contain the effect.
So very thanks to check it.


gaoyichuan:
Rejected due to poor code quality

ishitatsuyuki:
Test before opening: we're angry with such a PR

dog194:
-1
New item should be added until Genisys support 0.15.0

PR #1140, opened 2016-06-12
Fix for the /fill command

mad-hon:
Fixed parameter type checking, added progress reporting, added ability to use the command within console at least for default world


gaoyichuan:
Could you work on support relative parameters? such as use '~2' for the current position + 2? That will be alsome. Approved

PeratX:
+1

PR #1143, opened 2016-06-13
Fix events. Added PlayerBucketEmptyEvent

mixailenko5:


ishitatsuyuki:
Code Review -1: I think you're doing it opposite. Please test before posting.

jojoe77777:
I think that line should be left how it is, and the other PlayerBucketFillEvent should be changed to PlayerBucketEmptyEvent.

dktapps:
@jojoe77777 if you read the code carefully, it is in the right place. The code is admittedly very confusing to read though. `$targetBlock` is what's currently contained in the bucket.

ishitatsuyuki:
@dktapps What you are trying to explain are right but your conclusion is wrong. Read the code again.

dktapps:
@ishitatsuyuki as I said, the code is confusing. @jojoe77777 is right: the event calls are the wrong way round.

dktapps:
@ishitatsuyuki there seems to be an issue with pullapprove

ishitatsuyuki:
@dktapps Yes, I'm investigating.

ishitatsuyuki:
@dktapps I have finally fixed the pull approve. You can play something on this one, however please note that actions on your own PR will be ignored.

dktapps:
Including discussed changes in #1145 for the sake of simplicity.

PR #1147, opened 2016-06-13
Reverted array_filter function on args

Striker209:
Removed the array_filter call on args as it prevents players from teleporting to coordinates that contain zeros


dktapps:
Tested, looks good as far as I can tell.

ishitatsuyuki:
Code Review +1

PR #1159, opened 2016-06-14
Update Protocol 0.15

SuperKali:
Protocol 0.15


xxFlare:
@SuperKali This won't update the mechanics, just this allows clients to come (but it wont work) without proper changes to Login Packets.

Muqsit:
Oh, who even noticed the protocol had changed in 0.15?

daniktheboss:
when is the official 0.15 phar coming out?

Muqsit:
As soon as it could!

daniktheboss:
I read there is a 1.2dev version? But that one doesn't let people join. :(

PR #1179, opened 2016-06-16
0.15 Some Fixes

Muqsit:
- Fixed Leads on Players
- Temporarily removed additionalChar to avoid console spam

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -138,6 +138,7 @@ use pocketmine\network\protocol\PlayerActionPacket;
 use pocketmine\network\protocol\PlayStatusPacket;
 use pocketmine\network\protocol\RespawnPacket;
 use pocketmine\network\protocol\SetDifficultyPacket;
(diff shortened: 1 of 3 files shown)


dktapps:
I think `$additionalChar` and `getAdditionalChar()` can be safely removed. It is unused and PM doesn't have it anymore.

Muqsit:
But Steadfast2 uses it...maybe it is used somewhere..

dktapps:
I've searched the entire repo and there is no reference to it anywhere.

dktapps:
Uhm... leashes
image

Dog2puppy:
I think it looks good.

daniktheboss:
Did u fix it?

Muqsit:
@dktapps It happens with mobs but not with players.

dktapps:
I'm sorry to say it _does_ still affect players, I've tested it

Muqsit:
werw
There might be something else that the leashes get attracted to.

Muqsit:
Fine, whatever. I'll fix all at once.

PR #1183, opened 2016-06-16
Removed $additionalChar and getAdditionalChar()

Gamecrafter:
**1** file was modified. The removed lines are no longer needed, even PocketMine-MP itself doesn't have this.


Muqsit:
true

Gamecrafter:
@Muqsit How did you become a member? :P

dktapps:
Code Review +1

PR #1192, opened 2016-06-17
Fixed leads

legoboy0215:
Thank MiNET for this fix. I just implemented it into Genisys. NOT TESTED.


dktapps:
Amazing that you didn't see we already have an open PR with a fix for this that actually has been **tested** and works.

legoboy0215:
Oops.

Issue #1180, opened 2016-06-16
0.15 issue checklist - READ before opening a new issue!


dktapps:
As I'm sure you all know by now, Genisys has lots of issues with 0.15. Known issues are listed below.

- [x] [193a5c9] ~~Client-side crash when the player spawns in the vicinity of a mob or the player spawns a mob.~~

- [x] [193a5c9] ~~Players are connected to all mobs in the area (including other players) by leashes.~~ (not completely fixed)

- [ ] **Mobs in mob spawners have leashes attached which are constantly flailing around**

- [x] Various issues with custom mob plugins such as FNPC and Slapper, including leashes and client-side crashes. Apparently these issues have been resolved with updates to these plugins, please make sure your plugins are up to date.

- [x] [193a5c9] ~~Block updates are not sent to the client (explains issue #1178)~~

- [x] [4bf731c] ~~Batch block updates do not work properly. This causes issues with world editing plugins as only some of the changes are visible to clients unless they re-login.~~ (MCPE bug?)

- [x] [193a5c9] ~~Players' movements cannot be seen by other players. Actions such as swinging a sword are however visible.~~

- [x] [919a195] ~~Weather issues: rain/snow still happen even when weather is disabled in the configuration.~~

- [x] [7dd3afc] ~~Skins: Player skins always using the Alex model~~

- [x] Clients will crash in the vicinity of unknown blocks, i.e. blocks which are not currently in PE. This is an MCPE bug and as such is not the fault of Genisys.

- [ ] Enchanted arrows crash the client. Why arrows are able to be enchanted at all I'm not entirely sure, but this needs to be corrected.

Please comment below if you experience any issues not described here.

kabluinc:
when I spawn in my spawn, the floor is all covered in snow. after travelling some distance and coming back to spawn, the snow disappears and the land is back to normal. some biome/map issues? 

never happened with 0.14.3 and when the map is put in a 0.14 server there's no such issues

evolvedtm:
Although its a bit buggy, thank you for updating

dktapps:
@evolvedtm you mean a _lot_ buggy. We're working on it.

kaitododoyt:
weather doesn't work

evolvedtm:
Well, either way, I appreciate it @dktapps

kabluinc:
Weather has always been off from genisys.yml for my server but after update to 0.15 its snowing!? But on genisys.yml weather is still false

[...]

PR #1208, opened 2016-06-18
Fix player skin models always using Alex

dktapps:
Fixed player skins always using the Alex model (thanks @mhsjlw for finding the problem). Also reformatted login packet with proper indentation.
Another problem checked of #1180 .


Muqsit:
Nice one! +1

ishitatsuyuki:
I will approve it for now: seems good.

PR #1210, opened 2016-06-18
Update README.md

ScandiePlay:
I hope, my translation will be successful, 'cause I spent pretty much time, while doing it =D


ishitatsuyuki:
Sorry, but no thanks - English is enough, and Chinese is there because it's the developer's country.

PR #1214, opened 2016-06-19
just a little fix in this packet thing

Adam1609:


ishitatsuyuki:
You'd better go hospital.

PR #1224, opened 2016-06-20
Fixed CommandSelector Bug.

yuu528:


dktapps:
Looks okay, but please squash your commits by rebasing. This will be rejected until you do.

ishitatsuyuki:
Code review -1: this looks completely wrong, since @p will target the player itself unless there are criteria specified, which is unsupported.

dktapps:
Actually I think the question here is why the @p is searching for other players at all. @p should target the command sender, not nearby players.

PR #1226, opened 2016-06-20
Make @p command selector work the way it's supposed to

dktapps:
@p command selector should target the player.

### DIFF
--- a/src/pocketmine/command/SimpleCommandMap.php
+++ b/src/pocketmine/command/SimpleCommandMap.php
@@ -246,19 +246,11 @@ class SimpleCommandMap implements CommandMap{
 					break;
 				case "@p":
 					if($sender instanceof Player){
-						$distance = 5;
-						$nearestPlayer = $sender;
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
Code Review +1

gaoyichuan:
Code Review: Rejected
Reason: @p should select the nearest player, not the sender himself

ishitatsuyuki:
@gaoyichuan No, try it in PC.

dktapps:
When used by a command block, yes the nearest player is targeted. But if used by a player it will target that player unless used with arguments, which are not supported.

dktapps:
@gaoyichuan That approve hasn't been picked up, you might need to make a fresh comment

gaoyichuan:
Code review +1

JUSTaMod:
how do I use the selectors in in-game commands? Is that a thing? There are no cmd blocks.

gaoyichuan:
@JUSTaMod Just use them in-game or in the console.

Issue #1235, opened 2016-06-21
Hurt animation even though even is cancelled


legoboy0215:
#### Issue description
Title says it all.
#### Steps to reproduce the issue
Eh... Hit someone in spawn when you are using a plugin like PeacefulSpawn.
#### OS and versions
* Genisys: This server is running Genisys 1.2dev-919a195 "Ikaros" implementing API 2.0.0 for Minecraft: Pocket Edition v0.15.0.0 alpha (protocol version 81)
* PHP: 7.0.2
* OS: Linux Ubuntu

#### Crashdump, backtrace or other files
```
None
```
I really don't know the issue, but I see a patch 
[here](https://github.com/Hydreon/Steadfast2/commit/ccbf5c888e974dc2fa3084e6d3bfe88b1bcc6b29)

PS: In addition, there is this annoying death bug that happens to my server all the time, and another patch on Steadfast2 PROBABLY fixes it. The first issue is much more important though.
https://github.com/Hydreon/Steadfast2/commit/408baca649961a9be958cbe5926a87d29798c5c4

gaoyichuan:
Close for not following the template. If you continue doing this, you will get a ban

legoboy0215:
What?!

gaoyichuan:
Ah, ok. But I can't understand your language. What's

dktapps:
This annoying death bug. Is this the bug where no terrain loads when you respawn and the screen just kind of... jitters?

legoboy0215:
Yes.

jojoe77777:
Haven't tested this, but... https://github.com/NiclasOlofsson/MiNET/commit/6c9bc2bbee323cc2b8e676aec0265f6bc5151ba6

SleepSpace9:
Tested this https://github.com/ClearSkyTeam/ClearSky/commit/891b63ac4d038a3a6c2a8f445e487779885bed5e with Genisys and like @Guillaume351 said in Issue https://github.com/iTXTech/Genisys/issues/1180 it worked. So it would be nice to have also a commit at Genisys. ;-)

flaxues:
@SleepSpace9 Maybe you could make a pull request...

SleepSpace9:
Great idea, i'll try... :+1:

Muqsit:
#1237

PR #1237, opened 2016-06-21
Fix #1235

Muqsit:
Fix damage animation sent to client even after event gets cancelled!
Thanks to @tigerza117

PR #1238, opened 2016-06-21
Should stop "Hurt animation even though event is cancelled" (#1235)

SleepSpace9:
Fix for https://github.com/iTXTech/Genisys/issues/1235


dktapps:
Duplicate of #1237

SleepSpace9:
just seen...

PR #1134, opened 2016-06-11
Fixed the age-old hotbar glitch when changing slots rapidly

dktapps:
There is no need to send packets back to the client when they change hotbar slots. The server was returning several `MobEquipmentPacket`s to the client which were causing unnecessary hotbar slot changes.

I have tested this patch heavily before committing and it works perfectly to the best of my knowledge.

PR #1240, opened 2016-06-21
Fixed batch block updates using BatchPackets

dktapps:
Use `BatchPacket` and stuff all the single `BlockUpdatePacket`s inside it.
This saves DDoSing the client with thousands of single `BlockUpdatePacket`s as per some "fixes".

This restores original behaviour and compensates for the loss of functionality to send multiple block updates in a single packet. I also noticed that with these changes the client actually lags a lot less during large jobs such as world editing jobs.

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -912,15 +912,16 @@ class Level implements ChunkManager, Metadatable{
 	 * @param bool     $optimizeRebuilds
 	 */
 	public function sendBlocks(array $target, array $blocks, $flags = UpdateBlockPacket::FLAG_NONE, $optimizeRebuilds = false){
-		$pk = new UpdateBlockPacket();
-
+		$batchPacketList = [];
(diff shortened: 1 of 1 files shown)


Muqsit:
Tested, this fixed my magic carpet plugin <3 +1

markkrueg:
Testing this today. REALLY good. Feels like some dramatic improvements to lag/latency of actions. Even the hotbar seems improved with the batch packets.

Muqsit:
@markkrueg Speaking about the hotbar, it was fixed in this PR: https://github.com/iTXTech/Genisys/pull/1134

markkrueg:
@Muqsit Was that PR not merged into main 10 days ago? How is that PR related to this PR?

dktapps:
@markkrueg I can't imagine why batching produced such a big improvement, but I must admit I've noticed the difference myself.

gaoyichuan:
LGTM, approved

PR #1242, opened 2016-06-21
Change Golden Apple to PC recipe.

IrishUHC:
I am sure that using nuggets is intended, but is it possible to change it to the recipe on MC:PC?


gaoyichuan:
+1

Muqsit:
+1

PR #1260, opened 2016-06-24
Delete BanCidCommand.php

ABAxStorm:
It appears it does not have use.


gaoyichuan:
If you want to delete a command, you should at least delete the entry in SimpleCommandMap.php. and you will get a ban if you do such thing the next time

ABAxStorm:
Oh, sorry.

JUSTaMod:
Why would u get rid of Cid ban anyway? It's useful, as long as people don't know how to get around it. (It's a device ban. If you didn't know)

PR #1279, opened 2016-06-27
Changed Tile to BlockEntity (fully compatible with old APIs)

PeratX:


dktapps:
+1, tested and everything seems fine. But this is a huge API change.

Driesboy:
please not 
i need to update all the plugins
and players go left genisys

dktapps:
@Driesboy This update is **fully compatible with older APIs**. No changes need to be made to plugins.

Driesboy:
realy?

dktapps:
@Driesboy Yes. Test it if you don't believe me.

Driesboy:
ok i will test

PeratX:
Only for new plugins, I suggest use blockentity instead of tile.

gaoyichuan:
Tested and it worked. Approved.

PR #1288, opened 2016-06-28
Removed redundant code inside the FallingSand class

legoboy0215:
I don't get the point of that code. I made a plugin that spawned a FallingSand entity, and the entity was killed immediately after spawn if it was spawned mid-air. The issue was fixed after I removed this line of code.


dktapps:
Tested, code review +1

Muqsit:
+1

PR #1304, opened 2016-06-29
Respiration Enchantment!

dog194:
Respiration Enchantment is available now!

So up to now. Genisys support the enchantment effect below.
 - Tools: Efficiency,Silk Touch, Unbreaking, Fortune,
 - Armors: Protection, Fire Protection, Feather Falling, Blast Protection, Projectile Protection, Respiration, Aqua Affinity, Depth Strider, Unbreaking.
 - Sword: only Unbreaking. :(
 - Bow: null. :(

Other Enchantment effects are still on the way. XD


dog194:
! Notice!!!!: Do not wear a Helmet with Respiration in MCPE 0.15.0.1 now! It may causes a crash when you run out of breathing time. Blame mojang.

dktapps:
Nice! +1

Muqsit:
Wow +1

PR #1309, opened 2016-06-30
Fix some error (i didt test this code)

mixailenko5:


src/pocketmine/block/Cake.php
```diff
@@ -90,11 +90,11 @@ public function getDrops(Item $item) : array {
 	}
 
 	public function onActivate(Item $item, Player $player = null){
-		if($player instanceof Player and $player->getHealth() < $player->getMaxHealth()){
```
dog194: 0.0 . I think using $player->getMaxHealth() may be better.

dog194: oh. My fault.

PR #1314, opened 2016-06-30
Fixed critical bug in Wood2.php

Asidert:


dktapps:
Actually, I think you'll find that the same function is already defined in Wood.php, which Wood2 extends. This doesn't look to make any difference. Code Review -1.

Asidert:
With wood you can crash any server, checked on my own!

PR #1333, opened 2016-07-02
Troubleshooting $this->getInventory()->getItemInHand() and $this->inv…

tigerza117:
…entory->send Armor Contents ($player) on null in human temporarily.

It's hard for me But I do not recommend it. I'm just a temporary fix is not open, open minigame survival.


ishitatsuyuki:
No hack. Closing.

Driesboy:
client crash if join for the 2e x

PR #1338, opened 2016-07-02
Update Effect.php

biswarm22:


dktapps:
You'd better go get your eyes tested.

PR #1341, opened 2016-07-02
Fix Player not despawnFrom on timeout

tigerza117:


dktapps:
Code Review -1. Reason: When a player is kicked, the entity is automatically closed, so this will make no difference. Reference: https://github.com/iTXTech/Genisys/blob/master/src/pocketmine/Player.php#L3652

tigerza117:
This not work

PeratX:
Player will despawn when close

PR #1343, opened 2016-07-02
Added hasEnchantment.

IrishUHC:
This gets if the item has the specific enchantment stated as the parameter.


IrishUHC:
This could be the same as getEnchantment, but I'm not sure.

PeratX:
Your code is wrong! $enchantment is Enchantment object, but you tried to use it as int!

PR #1349, opened 2016-07-03
Add absorption effects?

XFuryMCPE:
Should work, will get back once I test it

### DIFF
--- a/src/pocketmine/entity/Effect.php
+++ b/src/pocketmine/entity/Effect.php
@@ -268,6 +268,11 @@ class Effect{
 					$entity->attack($ev->getFinalDamage(), $ev);
 				}
 				break;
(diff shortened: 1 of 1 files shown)


Muqsit:
Good jab! +1

XFuryMCPE:
It works? @Muqsit

Muqsit:
@XFuryMCPE I am not certain of that but Absorption wasn't registered in Effects completely.

dktapps:
Interesting. I'd prefer you test your code before making a PR though if possible, it saves all sorts of inconvenience.

XFuryMCPE:
@dktapps True, sorry about that. I tested it and nothing happens so I guess it didn't work :P dunno why..

PR #1368, opened 2016-07-05
Cyclic starting the server on Windows

samalero:


dktapps:
Code Review -1. This is copied directly from ImagicalMine and will only work if the server port is on 19132. Otherwise it has to be manually edited.

PR #1371, opened 2016-07-05
Cyclic restart of the server autodetection of port

samalero:
for Windows ;-)

### DIFF
--- a/start.cmd
+++ b/start.cmd
@@ -2,8 +2,23 @@
 TITLE Genisys server software for Minecraft: Pocket Edition
 cd /d %~dp0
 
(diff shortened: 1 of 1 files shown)


dktapps:
It would be preferable if you made this optional like the Linux start script.

samalero:
Well, today I will add ...

dktapps:
Code Review -1. Reason: This script will crash the first time the server is run.

dktapps:
No, this isn't the right way to do it. Now it starts lots and lots of server instances when the setup wizard is running.

PR #1397, opened 2016-07-07
Fix little code in EnchantedGoldenApple

KennanFatt:
Maybe @PeratX forget about it :D


Muqsit:
That won't make any difference.

PR #1400, opened 2016-07-08
Updated to 0.15.2

XFuryMCPE:
Not sure if the protocol actually needed to be updated.. Ppl on twitter were saying they couldn't join on 0.14.2 and this protocol change allowed me to, so yeah :P


Driesboy:
when it come out?

AndreTheGamer:
15.2 lets clients joins 15.x servers. I've tested it myself. And i was able to join any servers that were running on 15.x version(S).

Issue #23, opened 2015-12-30
I can't pvp other players :/


DWWf:
I doesn't let me pvp other people :/ just mobs and its hard to completely kill them

PeratX:
If it caused no damage,try ti turn off Hunger System.

DWWf:
How?

MCPEPIG:
genisys.yml

DWWf:
Tried it but still doesn't work.

astersx:
not genisys fault is a plugin u own!!!  Turn pvp on,  and a guess unuse ifriend+pureperms?

DWWf:
only pure perms

CrayonDing:
maybe some plugins cause this problem

PR #1429, opened 2016-07-11
Fix synapse issues #23

tigerza117:
fix freeze on tap the pause button (synapse only) #23


gaoyichuan:
This comment breaks PHPDoc

XFuryMCPE:
You can't just comment out a bunch of random crap to fix bugs... More bugs come from that. You need to find the actual cause of the bug and FIX it, not remove an entire important section of the source. Do you even know PHP?

PR #1440, opened 2016-07-12
Kill off duplication and make Block/Item classes so much neater

dktapps:
### Description
Clean up Block and Item classes and reduce duplication using interfaces containing the relevant block/item IDs. The access methods of these constants remains exactly the same as it did before.

### Reasons to modify
- Less mess
- Better consistency
- Less duplication

### Tests & Reviews
I have tested this and these changes do not affect or change functionality at all; all original functionality is retained. There are no compatibility issues with older APIs.


dschwartz783:
Is there any reason you'd want to implement BlockId or ItemId in any other class though?

src/pocketmine/item/ItemIds.php
```diff
@@ -0,0 +1,217 @@
...
+	const BEETROOT_SEEDS = 458; const BEETROOT_SEED = 458;
+	const BEETROOT_SOUP = 459;
+	const RAW_SALMON = 460;
+	const CLOWN_FISH = 461;
+	const PUFFER_FISH = 462;
+	const COOKED_SALMON = 463;
+
+	const ENCHANTED_GOLDEN_APPLE = 466;
+
+	const CAMERA = 498; #blamemojang
```
JUSTaMod: lol. blame mojang is right.

ishitatsuyuki:
I will leave you to choose: but expressing my opinion, Blocks shouldn't be mixed with Items. There are items not a block, blocks not an item. Items and Blocks are essentially different thing, just having many common IDs.

dschwartz783:
The logical separation is important to avoid confusion, Imo.

ZedCee:
This should improve code efficiency and maybe even increase speed a little by limiting the size of called namespaces for any given function, shouldn't it?

_PHP is a guessing game for me, but I would assume it decompresses the phar into a virtual environment of sorts.  If the phar is decompressed on-the-fly, then any improvement would only be evident running from src/..._

ZedCee:
@dktapps Haven't had enough time to really see and I couldn't imagine how to benchmark this: Start-up seems super snappy and I don't segfault on start anymore.
Stopping is another story, as it seems a segfault IS how it quits...   Wait, where are my core.:hankey:s?

Shit. 
After the 'rm -rf' incident (Damned jenkins.sh), I forgot to turn those off :worried: 
Oh dear...  /tmp...  
Gotta go

ishitatsuyuki:
No, this won't have a dramatic improvement. @ZedCee Please stop spamming with your noob knowledges.

dktapps:
@ishitatsuyuki that's why I split them up into two separate classes. I've made all Block constants accessible as Item constants also for the purposes of backwards compatibility.

ishitatsuyuki:
Well, if it retains the compatibility, this should be OK since it's the Mojang's design and it's just the ID part. Code review +1

PeratX:
+1

dktapps:
@ishitatsuyuki ...?

dktapps:
I'm done with that branch now

PR #1453, opened 2016-07-13
fix fence gate turning on activate

dschwartz783:
### Description
Fixes the fence gate turning to face the player every time it activates

### Reason to modify
Because the fence gate is not behaving correctly. It does behave correctly after this change.

### Tests & Reviews
I have tested this, and the fence gates no longer turn to face the player on activate, just on placement, as it should. Interestingly, the fence gate still always opens away from the player, which is also a good thing. It just doesn't turn if the player is facing the gate's side.

### DIFF
--- a/src/pocketmine/block/FenceGate.php
+++ b/src/pocketmine/block/FenceGate.php
@@ -110,8 +110,6 @@ class FenceGate extends Transparent implements ElectricalAppliance{
 			2 => 1,
 			3 => 2,
 		];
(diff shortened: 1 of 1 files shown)


dschwartz783:
To be honest, I wasn't sure if I needed to leave `$this->meta ^= 0x04;` What does that do exactly?

dschwartz783:
Not finished... Gate position is not remembered. I'll reopen it in a bit.

PR #1471, opened 2016-07-16
Fix #1456

### LINKED ISSUE #1456: Glowing Obsidian does not glow until local lighting is updated
#### Issue description
Glowing Obsidian only glows after a light source is added/or broken nearby.  A ominous glow emits around the blocks, but produces no light, nor do the blocks glow.

- Place Glowing Obsidian in a dark place.
- Place torch, watch your Obsidian glow.
- Remove torch, it will glow for you and any players that were nearby when activated until death or teleport.

#### OS and versions
* Genisys: 1.2dev-90c31fc
* PHP: 7.0.6
* Server OS: Linux
* Client OS: iOS and Android
* Game version: 0.15.2

img_8308

img_8309

XFuryMCPE:
### Description
Fix Glowing obsidian lighting

### Reason to modify
Fix #1456 

### Tests & Reviews
I have tested the code and it works. Left and rejoined server before change, glowing obsidian light wasn't updated and was left dark, now it lights up whenever teleported to it.

### DIFF
--- a/src/pocketmine/block/GlowingObsidian.php
+++ b/src/pocketmine/block/GlowingObsidian.php
@@ -22,7 +22,7 @@
 namespace pocketmine\block;
 
 
-class GlowingObsidian extends Solid{
+class GlowingObsidian extends Flowable{
 
 	protected $id = self::GLOWING_OBSIDIAN;
 


dktapps:
extends _Flowable_? Glowing Obsidian is a solid block. -1.

XFuryMCPE:
What about a torch.. A torch is a block.

Muqsit:
Shouldn't torch and itemframes extend transparent? If they're blocks, then if a player goes through a torch or an item frame, they'll experience (rubberband) pullbacks.

dktapps:
This was garbage. Not properly tested, and not properly _thought_ about. If you'd looked at the properties of Flowable before making your changes, you'd notice that Flowables have no bounding box, meaning they are **non-solid**, and therefore that that was a stupid idea. Result was that when you walked over it, anti-fly measures would be triggered because the client thought it didn't have a bounding box but the server did, meaning the server thought you were flying over a non-solid block.

https://github.com/iTXTech/Genisys/commit/b6b3aeb810633fc28709b2c10e84285e6f73a8b8 was the correct way to do this. In future **consider all possible results and implications of your changes** before opening a PR. Just because something "works" does not mean it's the best or even the right way of doing it.

PR #1480, opened 2016-07-17
Seems like it makes sense..?

XFuryMCPE:
### Description

Update tell command to show display name, rather than real name


ZedCee:
This breaks the tell command.  
As there really isn't anywhere we actually input a real name, I can't see how changing this is worth it or a good idea, and likely isn't part of the api to do so.

dktapps:
Some people never learn.

PR #1500, opened 2016-07-21
Change default format to anvil

dschwartz783:
### Description
Change genisys' default level format

### Reason to modify
MCRegion is not supported by some tools, such as MCEdit, and requires conversion for use in MCPC (MCPC is supposed to convert it automatically, but I've found that converting Genisys-created worlds is non-trivial, and requires a few steps).

### Tests & Reviews
Nothing to test really. Just a defaults change.

I'm proposing that we change the default format for new levels to Anvil, a superior format to MCRegion for a number of reasons, as well as the default format on PC, which will make things a lot easier to move back and forth, without conversion. I just find it's a far more sensible default format...

### DIFF
--- a/src/pocketmine/resources/pocketmine.yml
+++ b/src/pocketmine/resources/pocketmine.yml
@@ -92,7 +92,7 @@ debug:
 
 level-settings:
  #The default format that levels will use when created
- default-format: mcregion
+ default-format: anvil
  #If true, converts from a format that is not the default to the default format on load
  #NOTE: This is currently not implemented
  convert-format: false


dktapps:
Seems good to me. Code Review +1.

ishitatsuyuki:
I have heard that Anvil in PM is not stable before, is it still the case? If it's stable now, it's OK to switch.

dschwartz783:
The only issues I've seen are with unknown block types not being handled correctly by Genisys, which you'd expect to most often come with worlds in the Anvil format. That may be how that rumor got started. We've been using an Anvil-formatted world for awhile now with no issues.

ishitatsuyuki:
Then it seems OK. I approve this.

ZedCee:
I know it's a bit bit late, but... +1

dktapps:
The problem with unknown blocks is irrelevant of the world format. Hell, you can do it in singleplayer if you use mods.

dschwartz783:
@dktapps True, but unknown blocks are more common with Anvil because it's the default format for PC.

ZedCee:
@dschwartz783 Only if you're importing maps.  Blocks are just values.  You could likely put those same values into a mcregion and crash a client, as mcregion and anvils are basically the same base; Anvil just provides a few minor differences to enable taller worlds, better maintains biome data, and a better packet schema.
http://minecraft.gamepedia.com/Anvil_file_format

dschwartz783:
@ZedCee I knew all of that, and I know what the Anvil format is. I even began learning its structure because of an issue I was dealing with a little while back. I know that corrupt blocks can appear in any format, but what I said is still true. Because Anvil is the default format for PC worlds, that tends to be the format where you find invalid blocks. LevelDB should never have invalid blocks, and MCRegion should only contain blocks from old PC worlds, or worlds created by Genisys, unless the world was converted from Anvil. I like to think I know what I'm talking about. :)

ZedCee:
Ultimately it comes down to the map generation and thus imported maps have potential for invalid block types, but native maps should have no issues like those.

dschwartz783:
@ZedCee Correct.

PR #1509, opened 2016-07-23
SPP v5 and Login timeout

MrGenga:
### Description

Update to SPP v5 to work with Nemisys.

### Reason to modify

Because Nemisys is more stable, then Synapse.

### Tests & Reviews

I have tested that code, and it works. I tested it on my server. Works well, but my server is so laggy, because I have only 8GB RAM and 6 CPU cores with 15 started servers. Test server IP: pe.gc-game.ru(SkyWars and BedWars is on Nukkit with SynapseAPI). Now without stupid code lines, and normal code formatting. Without useless commits. Edited with git and nano on Ubuntu 16.04.

### DIFF
--- a/src/synapse/Synapse.php
+++ b/src/synapse/Synapse.php
@@ -24,6 +24,7 @@ namespace synapse;
 use pocketmine\Server;
 use pocketmine\utils\MainLogger;
 use pocketmine\utils\Utils;
+use synapse\network\protocol\spp\BroadcastPacket;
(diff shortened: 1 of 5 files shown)


dktapps:
Better. But do not spam PRs in future. This is your fourth one. Most people would receive a ban for such actions.

Driesboy:
how to instal nemisys on osx?

PR #1528, opened 2016-07-26
Add ChickenLayEggSound.php

XFuryMCPE:
### Description

Adds chicken lay egg sound

### Reason to modify

It sounds cool :laughing: 

### Tests & Reviews
Yeah it works.

### DIFF
--- a/src/pocketmine/level/sound/ChickenLayEggSound
+++ b/src/pocketmine/level/sound/ChickenLayEggSound
@@ -1,0 +1,29 @@
+<?php
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
Yeah you're idiot. No file extensions, and it won't make a sound unless you add some code to trigger it.

XFuryMCPE:
I wasn't trying to add it to the actual chicken, it _was_ so you can trigger it yourself.

ZedCee:
@XFuryMCPE Ask yourself:
- Without a game or user trigger method is this worth it on it's own?
- Could it be done with the plugin that would call it?
- Did I ask if it was worth committing as is?
- Does it follow the require naming schema?

Maybe some of this seems silly, but the goal of Genisys is to make a functional vanilla server without mod like additions.  Any modifications from the Minecraft idiom should likely be done with a plugin.  In the past, many great additions like the mob AI started as a plugin.

XFuryMCPE:
Gotcha @ZedCee , sorry

PR #1529, opened 2016-07-26
add invisible bedrock

dschwartz783:
### Description
Adds invisible bedrock support, to act as a barrier. Very useful in certain situations. Don't add this to creative items, because it's a sort of dangerous block. Hard to remove without manually removing it via console or plugin.

### Reason to modify
Because invisible bedrock doesn't work at the moment

### Tests & Reviews
I am now able to /give myself invisible bedrock, and place it. Removal is a problem, so this should only be obtainable through /give.

### DIFF
--- a/src/pocketmine/block/Block.php
+++ b/src/pocketmine/block/Block.php
@@ -304,6 +304,7 @@ class Block extends Position implements BlockIds, Metadatable{
 			self::$list[self::POWERED_REPEATER_BLOCK] = PoweredRepeater::class;
 			self::$list[self::UNPOWERED_REPEATER_BLOCK] = UnpoweredRepeater::class;
 			self::$list[self::CAULDRON_BLOCK] = Cauldron::class;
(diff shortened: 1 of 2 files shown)


Driesboy:
Nice!

dschwartz783:
I thought it might come in handy, so I designed a tiny plugin to remove these blocks (since you can't remove them normally), which can be cloned here: https://github.com/dschwartz783/PMInvisibleBedrockRemover

dktapps:
I'd like to point out that Invsible Bedrock is a _transparent_ block, and therefore should _not_ extend Solid or this will result in lighting glitches. Rejecting until this is corrected.

dschwartz783:
Fixed.

dktapps:
All right, seems good. Code Review +1

ZedCee:
+1

dog194:
+1

[Approved with PullApprove](https://pullapprove.com/iTXTech/Genisys/pull-request/1529/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=iTXTech/Genisys)

PR #1531, opened 2016-07-26
New getter called getLevel();

sero583:
### Description

For a new function.

### Reason to modify

It is a good way in my opinon and i don't think it will do anything.

### Tests & Reviews

I didn't, but i'm to 99% sure it would work

New getter function, where you can get the level.


dktapps:
Code Review -1. This already exists in its parent class LevelEvent and is therefore pointless. Untested PRs are also unwelcome as they make extra work for the devs.

PR #1533, opened 2016-07-26
Fix crash MCPE on incorrect block

MrGenga:
### Description

Fix crashes of MCPE client, on receive unknown block(bug in 0.15).

### Reason to modify

I trying to fix crash on unknown blocks. There is PC maps, yeah!

### Tests & Reviews

There is no visual effect, only replaces unknown blocks to UPDATE block, like in Singleplayer.


dktapps:
Code Review -1. This will only prevent **updated** unknown blocks being sent to the client.

PR #1553, opened 2016-07-28
0.15.4 Protocol Fix

TheAppleGamer:
### Description

Part one of 0.15.4

### Reason to modify

Players can join


ishitatsuyuki:
We don't need such a PR unless you're sure everything else is also working correctly. Have you tested this?

TheAppleGamer:
I've tested it and it works for MCPE 0.15.0 - 0.15.4

ishitatsuyuki:
Do not use GitHub only to edit files. Please squash the commits.

TheAppleGamer:
Don't know github that well so didn't understand that one bit, and that's all that is needed for 0.15.4

PR #1559, opened 2016-07-29
Add IP message on server startup

ZedCee:
### Description

Adds the external and internal ip found in the installer to the main start up of the server.  

### Reason to modify

This helps with monitoring any changes to your IP every time the server is restarted, as opposed to the one-off displayed in the installer.

### Tests & Reviews

The remaining lang files require an additional line (pocketmine.server.ip_warning).


dog194:
0.0 maybe add pocketmine.server.ip_warning for all language files would be bettle(Also can use English temporarily) just a suggestion.

ZedCee:
@dog194 A work in progress.  I'm on it right now actually, just gonna take me a little bit

ZedCee:
Suffering a git error.  Just need to clean up. I will be back once I solve my new issue.

ZedCee:
Commits are squashed, not sure why the merge keeps happening, but probably user error.  Still learning the syntax.

PeratX:
This will slow down the startup sequence.

ZedCee:
@PeratX Personally, I don't find the extra few seconds to be a huge issue for myself.  
It allows me to see what my users may not report (I don't have very many users/testers and they'd rather play single player than report issues without my preemptive intervention) and saves time in the long run. 

I'll look into async tasks tonight (and see what I can come up with tomorrow; it just turned midnight here) but that may be a little beyond my realm of PHP knowledge, so anyone is welcome to contribute if they understand the syntax a better.

ishitatsuyuki:
Beware, it's a large slow down. You may need to change to a faster service.
By the way, this may break LAN servers without network.

ZedCee:
@ishitatsuyuki  I don't disagree that the 5-10 second delay may be considered large by some.  The issue is that ISPs randomize the IP at random intervals and ip_warning left only in the installer leaves many using other external sources more often than not.  This renders the ip_warning function only useful once upon setup.  Perhaps a switch for this would be more viable?  I agree async would be nice.

Without a network it actually starts quicker and I can't seem to produce any breakages, but not all systems and setups are the same.

src/pocketmine/lang/locale/jpn.ini
```diff
@@ -227,6 +227,7 @@ pocketmine.server.start=Minecraft: PEサーバー({%0}に対応)を起動して
 pocketmine.server.networkError=[Network] {%1}によって{%0}のインターフェイスが停止しました
 pocketmine.server.networkStart={%0}:{%1}上でサーバーを開始しています
 pocketmine.server.info=このサーバーは{%0}のバージョン{%1}「{%2}」(API {%3})で動作しています
+pocketmine.serve.ip_warning = あなたのグローバルIPは{%0}です。プライベートIP{%1}をポート解放してください。
```
ishitatsuyuki: ~~をポート解放~~ **にポート解放**

ishitatsuyuki:
The current implementation relying on DynDNS is a bottleneck. Please change the code to use a faster and ease-to-parse API: https://api.ipify.org

dktapps:
Rebased and utils is now using api.ipify.org

ZedCee:
I also attempted implementing api.ipify.org with mixed results.

@dktapps' runs better than mine, utils didn't like an extra else.  GG

@ishitatsuyuki Like without parsing? :confused:

PeratX:
Use AsyncTask

ishitatsuyuki:
Things to fix:
1. JP Translation
2. Error checking instead of regex

ishitatsuyuki:
@ZedCee after that, you can use git rebase -i --autosquash. This would be better with fix-up though, which discards the duplicate commit messages.

ZedCee:
@ishitatsuyuki Finally figuring out my mistakes, it was all in the ```push```.   Thanks.  ```--autosquash``` is nice to learn about, as it fills out what you need from the start and saves some work.  

@PeratX Is AsyncTask really necessary?  After the change to api.ipfy.org the bottleneck is imperceptible.   Virtually zero difference to the start up time, with or without ip-warning. 

Going to look into error checking now...  Let's see if this *PHP Luddite* can figure it out

dktapps:
I'll have a tinker with it later and get rid of the regex. AsyncTask is overkill.

ishitatsuyuki:
Seems the ipify still takes 1 sec on my environment. I will cherry-pick the change of the server, but not the startup message. The current startup time is 0.3s, which 1s is a big addition to that.

ZedCee:
Alright, I'm going to close this out and look for a better way to implement it.
As @dktapps suggested on gitter.im ip-warning, may actually be more useful as a default /command.

PR #1567, opened 2016-07-30
Added Czech/Čeština Lanauge

ondraczsk:
Added ces.ini


gaoyichuan:
Did you translate it by hand? If so, we would be appreciated.

dktapps:
If you could also translate the installer, that would be good.

ondraczsk:
ok

PeratX:
+1

dog194:
+1

PR #1570, opened 2016-07-31
Big weather improvements and cleanup. Fixed #1472

### LINKED ISSUE #1472: Clients do not see the same weather effects
#### Issue description
Clients do not see the same weather when they are in the same world. The below screenshot shows two players in the same world at the same position, looking in the same direction. (Left is PE on Nox, right is Win10 Edition)

Not entirely sure how to reproduce this. Will update if I figure out how.

#### OS and versions

* Genisys: b6b3aeb, 63769e6
* PHP: 7.0.8
* Server OS: Windows 10 Home x64
* Game version: MCPE and MC Win10 (both 0.15.2)

image

dktapps:
### Description
Fixes lots of annoying weather bugs, and general cleanup and refinement of weather manager

### Reason to modify
- Fix #1472 - clients who join after the weather changes to rainy would sometimes not receive weather updates
- Fix bugs such as ALWAYS raining/stormy
- Less packet spamming (there's no need to send 4 level event packets for rain/thunder change)
- Less code duplication, more consistency and cleaner code.
- Send a usage message when using /weather without the right parameters in the console instead of saying "Wrong parameters" (such a useless error)

### Tests & Reviews
I have tested the code and everything seems to work flawlessly.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2293,6 +2293,8 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 		$pk->metadata = [self::DATA_LEAD_HOLDER => [self::DATA_TYPE_LONG, -1]];
 		$this->dataPacket($pk);
 		
+		$this->level->getWeather()->sendWeather($this);
+		
 		$this->forceMovement = $this->teleportPosition = $this->getPosition();
 	}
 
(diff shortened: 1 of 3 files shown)


PeratX:
+1

ishitatsuyuki:
Good. +1

PR #1572, opened 2016-08-01
Add support for minecart to follow the rails

gimilii:
### Description
Added support where minecarts will now follow the rails (when the associated option is set in the genisys.yml file).

### Reason to modify
Currently the minecart will move freely, but it does not follow the rail  I had worked on getting a minecart to follow the rails several months ago on pocketmine, but I wasn't able to figure out how to have the player actually get in the cart (so good job to who ever figured that out).  I have now taken the code I had before for following the rail and ported it to the genisys code (never ended up committing it to pocketmine).

### Tests & Reviews

I have tested the code and it works.  I have tried various tracks (curves, up / down), next to blocks, etc.  It seems to work quite well.  There are still some things I would like to improve, but this seemed like a good time to create the pull request as it is functional.

Please review things below:
I am not sure though how to know when a user touches the forward button after entering the cart.  So for the moment the cart starts moving once the player gets in if it is on rail.

### DIFF
--- a/src/pocketmine/entity/Minecart.php
+++ b/src/pocketmine/entity/Minecart.php
@@ -52,7 +52,7 @@ class Minecart extends Vehicle{
 	public $isMoving = false;
 	public $moveSpeed = 0.4;
 	
-	private $onRail = Minecart::STATE_INITIAL;
+	private $state = Minecart::STATE_INITIAL;
 	private $direction = -1;
 	private $moveVector = [];
 	private $requestedPosition = null;
@@ -60,10 +60,6 @@ class Minecart extends Vehicle{
 	public function initEntity(){
 		$this->setMaxHealth(1);
 		$this->setHealth($this->getMaxHealth());
-		$this->moveVector[Entity::NORTH] = new Vector3(-1, 0, 0);
-		$this->moveVector[Entity::SOUTH] = new Vector3(1, 0, 0);
-		$this->moveVector[Entity::EAST] = new Vector3(0, 0, -1);
-		$this->moveVector[Entity::WEST] = new Vector3(0, 0, 1);
 		$this->moveVector[Entity::NORTH] = new Vector3(-1, 0, 0);
 		$this->moveVector[Entity::SOUTH] = new Vector3(1, 0, 0);
 		$this->moveVector[Entity::EAST] = new Vector3(0, 0, -1);
@@ -153,14 +149,39 @@ class Minecart extends Vehicle{
 	 * Check if minecart is currently on a rail and if so center the cart.
 	 */
 	private function checkIfOnRail() {
-		$rail = $this->level->getBlock($this->temporalVector->setComponents($this->x, $this->y - 1, $this->z));
-		if ($rail != null && in_array($rail->getId(), [Block::RAIL, Block::ACTIVATOR_RAIL, Block::DETECTOR_RAIL, Block::POWERED_RAIL])) {
-			$rail = $rail->floor()->add(.5, 0, .5); 
-			$this->setPosition($rail);    // Move minecart to center of rail
-			$this->state = Minecart::STATE_ON_RAIL;
-		} else {
+		for ($y = -1; $y !=2 && $this->state == Minecart::STATE_INITIAL; $y++) {
+			$positionToCheck = $this->temporalVector->setComponents($this->x, $this->y + $y, $this->z);
+			$block = $this->level->getBlock($positionToCheck);
+			if ($this->isRail($block)) {
+				$minecartPosition = $positionToCheck->floor()->add(.5, 0, .5);
+				$this->setPosition($minecartPosition);    // Move minecart to center of rail
+				$this->state = Minecart::STATE_ON_RAIL;
+			}
+		}
+		if ($this->state != Minecart::STATE_ON_RAIL) {
 			$this->state = Minecart::STATE_OFF_RAIL;
 		}
+	}
+	
+	private function isRail($rail) {
+		if ($rail != null && in_array($rail->getId(), [Block::RAIL, Block::ACTIVATOR_RAIL, Block::DETECTOR_RAIL, Block::POWERED_RAIL])) {
+			return true;
+		}
+		return false;
+	}
+	
+	private function getCurrentRail() {
+		$block = $this->getLevel()->getBlock($this);
+		if ($this->isRail($block)) {
+			return $block;
+		} 
+		// Rail could be one block below descending down
+		$down = $this->temporalVector->setComponents($this->x, $this->y - 1, $this->z);
+		$block = $this->getLevel()->getBlock($down);
+		if ($this->isRail($block)) {
+			return $block;
+		}
+		return null;
 	}
 	
 	/**
@@ -175,14 +196,24 @@ class Minecart extends Vehicle{
 		} else {
 			$candidateDirection = $this->direction;
 		}
-		$block = $this->getLevel()->getBlock($this);
-		$railType = $block->getDamage ();
-		$nextDirection = $this->getDirectionToMove($railType, $candidateDirection);
-		if ($nextDirection != -1) {
-			$this->direction = $nextDirection;
-			return $this->moveIfRail();
+		$rail = $this->getCurrentRail();
+		if ($rail != null) {
+			$railType = $rail->getDamage ();
+			$nextDirection = $this->getDirectionToMove($railType, $candidateDirection);
+			if ($nextDirection != -1) {
+				$this->direction = $nextDirection;
+				$moved = $this->checkForVertical($railType, $nextDirection);
+				if (!$moved) {
+					return $this->moveIfRail();
+				} else {
+					return true;
+				}
+			} else {
+				$this->direction = -1;  // Was not able to determine direction to move, so wait for player to look in valid direction
+			}
 		} else {
-			$this->direction = -1;  // Was not able to determine direction to move, so wait for player to look in valid direction
+			// Not able to find rail
+			$this->state = Minecart::STATE_INITIAL;
 		}
 		return false;
 	}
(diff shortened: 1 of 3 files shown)


Yoyu666:
wow

src/pocketmine/entity/Minecart.php
```diff
@@ -110,37 +133,258 @@ public function onUpdate($currentTick){
...
+					if ($this->state == Minecart::STATE_ON_RAIL) {
+						$hasUpdate = $this->forwardOnRail($p);
 						$this->updateMovement();
 					}
 				}
-				$hasUpdate = true;
 			}
 		}
+		
+		$this->timings->stopTiming ();
```
ishitatsuyuki: Some formatting issue.

ishitatsuyuki:
Nice work. Just a few point to ask:
1. Does this has speed calculation?
2. Can this go through curved unconnected rails by placing an accelerator rail?
```
 |
_/
 A
```

PeratX:
Nice

src/pocketmine/entity/Minecart.php
```diff
@@ -37,19 +38,36 @@ class Minecart extends Vehicle{
...
+	public $width = 0.98;
 
 	public $drag = 0.1;
 	public $gravity = 0.5;
 
 	public $isMoving = false;
 	public $moveSpeed = 0.4;
-
+	
+	private $onRail = MineCart::STATE_INITIAL;
```
dktapps: `MineCart`: please correct this to the right case.

robske110:
nice !

robske110:
@ishitatsuyuki 
To 1: No, seems to be hardcoded to 0.4

src/pocketmine/entity/Minecart.php
gimilii: I would be happy to fix.  Are you saying that STATE_INITIAL should be StateInitial?  I was following what I saw in code for other constants like TYPE_HOPPER

robske110: No he means MineCart should be Minecart

gimilii:
To 1: and comment on 1.  Right now it is using the pre-existing $moveSpeed which was set to a value of 0.4.  I have tried other speeds as well, values between x > 0 x <=1 seem to work fine.  I did notice an issue with Minecraft PE app freezing if going too fast, and so I added code to check the current tick and last update.  So at the moment the cart goes as a constant pre-defined speed.  This can be changed to vary base on rail type, and I would also like to figure out how to detect the user pressing the forward button when in cart so the cart wouldn't move on its own on non powered rail.

As to question regarding curved unconnected rails and accelerator rail, I am not sure I understand the question.  It does go through curved rails and up/down rails, but at the moment treats all rail (powered/non powered) the same.

src/pocketmine/entity/Minecart.php
gimilii: Removed blank line.

src/pocketmine/entity/Minecart.php
gimilii: Ah, thank you!  Fixed.

dktapps:
Well, I must say this looks amazing. I'll test this and see how it goes. Also, please squash your commits by rebasing if you can.

robske110:
You seem to always use setPosition? Isn't this resulting in non-smooth movement?

gimilii:
I did a rebase/squash - I hope I did it correctly - I am somewhat new to this environment.  Regarding setPosition.  I only use in two places.   1) checkIfOnRail - initially ensuring that the minecart is centered on the rail.   2) moveUsingVector special case logic.  I hit a case where the minecart does not move to the requested location, I think due to the fact that it would overlap with another block.  This occurs when moving up an ascending rail.  I first tried to address by adding 1 to dy, but that didn't work.  Forcing the position using setPosition did work.  I am honestly not happy with it, and was hoping others could tell me how to do this correctly.   Until then though this addressed the issue, and is only called in certain exception cases.

dktapps:
@gimilii looks fine, good job.

gimilii:
Oh, also thank you everyone for the nice comments!  And thank you for all your work on Genisys, we have just recently started using it, and the features are awesome!

robske110:
@gimilii I'd like to ask if i can implement this PR in CS, too. Of course with proper acknowledgement.

gimilii:
@robske110 Sure, I have no issue with that.

gaoyichuan:
Seems wonderful! I'll test this when I am free

robske110:
And @gimlii just in case you didn't know setMotion(0,1,0) doesn't move forward 1 Block. You need to calculate setMotion() parameters in other ways for that. I'll pull request when i'm done with that.

dktapps:
Important issues off-the-hop:
- movement-type 0 results in jumping bouncy minecarts
- Critical Undefined property $state on line 136 in Minecart.php
- Minecarts always get stuck going downhill in one specific direction (I think North).

Less important, but still weird:
- Players vibrate when in a moving minecart

Please try and correct these issues.

src/pocketmine/entity/Minecart.php
```diff
@@ -37,19 +38,36 @@ class Minecart extends Vehicle{
...
 		$this->setMaxHealth(1);
 		$this->setHealth($this->getMaxHealth());
+		$this->moveVector[Entity::NORTH] = new Vector3(-1, 0, 0);
+		$this->moveVector[Entity::SOUTH] = new Vector3(1, 0, 0);
+		$this->moveVector[Entity::EAST] = new Vector3(0, 0, -1);
+		$this->moveVector[Entity::WEST] = new Vector3(0, 0, 1);
+		$this->moveVector[Entity::NORTH] = new Vector3(-1, 0, 0);
+		$this->moveVector[Entity::SOUTH] = new Vector3(1, 0, 0);
+		$this->moveVector[Entity::EAST] = new Vector3(0, 0, -1);
+		$this->moveVector[Entity::WEST] = new Vector3(0, 0, 1);
```
dktapps: Why are these duplicated?

robske110:
to the $state issue: just rename (line55) $onRail to $state

gimilii:
@robske110 I don't think I use setMotion.  I do use setPositon where I add one to the current y value.  This was to handle the case of moving up when stuck.  If though its wrong and you correct please let me know details as I am still learning this code.  Thank you.

src/pocketmine/entity/Minecart.php
gimilii: Sorry, thought I had cleaned things up, but missed this.  I will fix.   At one point I was going to have UpNorth, ..., DownNorth, ... but then didn't and forgot to remove the copied lines.

gimilii:
@dktapps When you say "movement-type 0 results in jumping bouncy minecarts" can you explain in more detail?  What is movement type 0?  How do I recreate?

I will fix the undefined property issue.

I will test with mine carts going north and down.  I thought I had tested all directions but looks like I missed one.  To be honest now that I think of it, I had rail going in an oval with it going up and down on two sides, and I think both of those sides were the West-East sides.  I am probably just not handling right in the North South.

Regarding players vibrating, does this always happen, or in certain situations.  I didn't notice this when I tested, and I was able to look around freely.  I did notice if the speed was too fast (both cart speed and update rate) that then things would start to freeze.

dktapps:
@gimilii From a 3rd-person point of view, players vibrate. And when minecart movement type is set to 0 in genisys.yml, minecarts will bounce up and down on rails.

NycuRO:
Thx salmon for reclam :))))))))))))) Srry man,idk is from other person :p

caspervanneck:
@dktapps i dont se my Players vibrate

robske110:
@dktapps means, that AN OTHER player sees AN OTHER player vibrating.
and NycuRO stop breaking licenses.

dktapps:
@NycuRO @caspervanneck Trying to claim credit for somebody else's work can be classed as a felony. What you did there is technically illegal, ESPECIALLY since you copied @gimilii 's work, tried to take credit for it AND tried to place it under your proprietary "licence".

caspervanneck:
@dktapps i say tot him thats a pull

caspervanneck:
PS i dont work any more for apollo

caspervanneck:
End im sorry wat happend

robske110:
@dktapps and he just blindly copies my commits. It is a fork of CS. and sometimes he copies stuff so bad that it doesn't even work anymore. and the mine cart patch isn't even implemented right yet

gimilii:
@dktapps Regarding "movement-type 0 results in jumping bouncy minecarts"  I hadn't tested with movement-type 0 since I was adding support for movement-type 1.  Its actually kind of funny to see the minecart jump.  Anyways, as far as I can tell this isn't due to my code change as I put back the original phar (before my modifications) and I notice the same behavior when a minecart is placed on a rail if movement-type is 0.  I can look at figuring out why it does that and fix it though if you like.  First I will look into the other issues reported.

caspervanneck:
You can ban mij @robske110 end @dktapps

caspervanneck:
Im sorry ):

src/pocketmine/entity/Minecart.php
```diff
@@ -37,19 +38,36 @@ class Minecart extends Vehicle{
...
+	public $width = 0.98;
 
 	public $drag = 0.1;
 	public $gravity = 0.5;
 
 	public $isMoving = false;
 	public $moveSpeed = 0.4;
-
+	
+	private $onRail = Minecart::STATE_INITIAL;
```
robske110: Simply rename this to $state

dktapps:
@gimilii me and @robske110 got a right laugh from that, yeah :D

NycuRO:
Dude @dktapps  i not breaking license.
Generator and updater is official by Apollo. Generator is modificated lot and not even with the looks to genisys and updater is from SalmonDE.

I'm saying NOT copy generator and updater,not all files ;)

And @robske110 you say i copy your commit ,but you and dan copy all generator from genisys ,dude. :/

robske110:
we have no generator from genisys. AND YOU MUST GIVE FUCKING CREDIT. IF YOU DO NOT GIVE CREDIT YOU ARE BREAKING THE LICENSE AND THEREFORE DOING STH  I L L E G A L.

dktapps:
This is a PR discussion. Enough. Deleting all comments not relevant to this PR.

robske110:
thanks.
Anyways, back to this PR.
I'm now actively improving this @ ClearSky
stay tuned for commits there :)

gimilii:
@dktapp I am looking into issue of carts getting stuck, haven't completely figured out yet.  I think it is still related to the bounding box.
@robske110 Let me know if you see issues / fix, I can make the same changes in this pull request (with credit).

gimilii:
I have figured out the issue of cart getting stuck.  There were two problems.  1) My code to keep the cart centered wasn't working allowing the cart to drift to one side or the other and get blocked by nearby walls - I have addressed that.  2) When going up an ascending rail, once the cart gets half way going any further would cause it to move into the block underneath the rail above, so the cart is blocked.  I am still not sure of the best solution but at the moment I have added code so that at that point it moves on to the next block and up.  I probably need to refine this, so as to have the correct speed while going up, but truth is that will need work at some point anyway to handle deceleration and acceleration.  I need to do cleanup and then will commit my changes today.

gimilii:
I have made multiple changes as I figured out more regarding how the minecart needs to travel.  I have done a lot of testing, and the minecart should move now without getting stuck (I hope).  Now that I understand things better, I may want to rework this code, but wanted to finish this pull request first.  Regarding the player vibrating, I dont think that is related to my new code (minecart-moving-type: 1) as I see the same thing when the type is set at zero.  I looked at it a little and it seems to me to be related to the code that links the entity, and the fact that other clients seem to think the player is in two places.  I also noticed that when exiting the cart, other players still see a player in the cart.  I am guessing that maybe the Player code needs special logic for when there is a linked entity.  Not sure if anyone else would have a better idea of how that should work?  Once this pull request is approved (assuming it is) I can work on some refactoring, which would include allowing the cart to keep going (but slowing down) if it moves onto a square without rail.

I also played with two other modes - one like moving-type 0, but where it will be blocked by blocks in its way, thought this could be used for a mine cart racing type of thing.  And another mode where the mine cart flies in the direction you are looking as well as traveling on the ground.  I didn't commit those changes though.

PeratX:
Nice work.

dktapps:
@gimilii I'm going to cherry-pick the contents of this PR into a branch where we can modify it more freely. If you have further changes to make after that, feel free to make a pull request to that branch. Once again, thanks for your work :D

dktapps:
Okay, the branch is here: https://github.com/iTXTech/Genisys/tree/minecarts

gimilii:
@dktapps Is there a reason this can't be merged into master?

dktapps:
@gimilii The behaviour isn't yet vanilla and there are a few things that I think need ironing out. Once this is done it'll be merged into master.

gimilii:
@dktapps (Sorry for this long post).  In all fairness the original minecart behavior also isn't vanilla (nor are a lot of other things) but they are merged into master.  I think it was good that the code that allowed the minecart to move with a player in it was merged to master.  That is what triggered me adding the mine cart on rail support I had started awhile back but had abandoned due to having issues linking the entities.  The minecart on rails is not perfect, but I think it is an improvement over what is currently in the master branch.  Its presence may cause others to make further improvements as the ability to get in the minecart prompted me.

The issues (cart going up and down, and player flashing) I think are not caused by the pull request but are pre-existing.  I have not looked into the cart moving up and down yet, but I do think I understand the problem with the player flashing.

The player flashing involves several issues with the current master code branch.  The main issue seems to be that when a player enters the minecart, the player is both moved with the minecart (as they are linked), but the player is also still moved based on the client sending move player packets.  Other clients then are told of the player at two different locations.  This is because when the server tells the client of the player in the minecart to move, the client sends several move player packets, as it slowly moves the player to that location.  The client is always behind where the server thinks the player (in minecart) is.  I changed to code to not move the minecart until the player has caught up (based on the move player packets) and the flashing then is only vertical.  That is due to the fact that the client reports a different Y value (maybe due to sitting down, not sitting down).  If I tweak that value as well, I can get the player to not flash at all, but the cart moves very slowly as it has to wait for the client packets to catch the player up.

The other option is to just ignore the move player packets from the client when a player is linked to a vehicle.  This gets rid of the flashing completely and allows the minecart to move quickly.  The downside is that then the player is never seen to look around, even if the player is turning around since the move player packet includes pitch and yaw as well as the position.

My best thought now is to ignore the packets unless pitch and yaw are changing, and if they change, only apply that, but that still seems wrong to me.  I was going to explain this in a different pull request and see if others had ideas of the best way to handle.  Or maybe there is a different forum/mechanism I should use for that?

There are other issues I found though with the linking of the player and mine cart.  I am not sure, but I think the code that links/unlinks may not be quite right.  When it unlinks it unlinks in both directions, but links only in one.  I modified it to link in both hoping that would resolve the flashing issue (before I understood it), but it didn't.  It could be causing other issues though, or maybe linkage is only required in one direction, I am not sure.   One thing that does seem to be an issue is if you are already linked and then another player joins.  When a player links, a broadcast packet is sent to let the other players know of the linkage, and so their UIs show things correctly.  If though a player joins after, they don't know there is a link, and so they see the player running after the cart.  I will say this is what I saw on my branch of the code, I did not go back and reverify on master, so can not say 100% my code didn't cause this some how, I can try to verify that today or tomorrow.

Anyway, I think that this pull request is already functional, and improves the support for minecart in Genisys, and would like to see if merged to master.  Then I could create a separate pull request to try to address the flashing, but may need some help with that.  I partly want this as I want to continue to pick up Genisys updates, and don't want to have to re-merge my changes in each time.  I also have a goal to actually commit something to an open source project, and thought I would achieve it with this change (-:

dktapps:
@gimilii Please, bring this to gitter.im/itxtech/genisys or mattermost.itxtech.org

ZedCee:
@gimilii You have nailed your goal.  Don't worry.  We just want to separate the previous bugs from the new bugs and make sure that this is the absolute best it can be before a merge.  
A branch will give us the opportunity to help with the work and ensure code integrity, as well as tackle anything new that may have come about, or old the impedes your work.  
Come over to the networks mentioned by @dktapps and we can discuss and develop further.

PR #1576, opened 2016-08-01
Fixed spectators

Tarik02:
### Description
Now arrows don't affect spectators, players can build in spectators.

### Reason to modify
In many mini-games, spectators can prevent players to play.

### Tests & Reviews
It's was tested.

### DIFF
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -760,7 +760,7 @@ abstract class Entity extends Location implements Metadatable{
 	}
 
 	public function canCollideWith(Entity $entity){
-		return !$this->justCreated and $entity !== $this;
+		return !$this->justCreated and $entity !== $this and !($entity instanceof Player and !$entity->isSpectator());
 	}
 
 	protected function checkObstruction($x, $y, $z){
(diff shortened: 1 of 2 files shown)


PeratX:
+1

Driesboy:
+1 approved

src/pocketmine/entity/Entity.php
```diff
@@ -760,7 +760,7 @@ public function setMaxHealth($amount){
 	}
 
 	public function canCollideWith(Entity $entity){
-		return !$this->justCreated and $entity !== $this;
+		return !$this->justCreated and $entity !== $this and !($entity instanceof Player and !$entity->isSpectator());
```
dktapps: I think your logic here is incorrect.

robske110: Translating code to english:
when the entity (this) is not justCreated and the $entity isn't this entity and if the entity is not a instance of a player and is not a spectator the
So the entity cannot be a player wich is in spectator ....

dktapps: Yes, that's what I thought.

dktapps:
I'm going to close this, because there is a cleaner solution and this solution does not work correctly. Thank you for trying though.

PR #1587, opened 2016-08-02
Fix #1574

### LINKED ISSUE #1574: Bow cannot be shoot when jump once
#### Issue description

Bow cannot be shoot when jump once

Hold the screen to activate the bow, jump once, and the arrow won't shoot.
The same thing is not happened when not jumping.

#### OS and versions

* Genisys: 37abb4d
* PHP: 7.0.3
* Server OS: Windows 10
* Game version: Minecraft Windows 10 Edition

```
```

svilex:
### Reason to modify
Issue #1574

### Tests & Reviews

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2791,10 +2791,10 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 						}else{
 							$thrownPotion->spawnToAll();
 						}
+					}elseif($item->getId() == Item::BOW){
+						$this->setDataFlag(self::DATA_FLAGS, self::DATA_FLAG_ACTION, true);
+						$this->startAction = $this->server->getTick();
 					}
-
-					$this->setDataFlag(self::DATA_FLAGS, self::DATA_FLAG_ACTION, true);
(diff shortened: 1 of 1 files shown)


dktapps:
And you've DEFINITELY 100% tested this, yes? As I recall, your last 2 PRs broke lost of things because they weren't properly tested.

src/pocketmine/Player.php
```diff
@@ -2791,10 +2791,10 @@ public function handleDataPacket(DataPacket $packet){
 						}else{
 							$thrownPotion->spawnToAll();
 						}
+					}elseif($item->getId() == Item::BOW){
+						$this->setDataFlag(self::DATA_FLAGS, self::DATA_FLAG_ACTION, true);
+						$this->startAction = $this->server->getTick();
 					}
-
-					$this->setDataFlag(self::DATA_FLAGS, self::DATA_FLAG_ACTION, true);
-					$this->startAction = $this->server->getTick();
```
dktapps: I'd like to observe that this is more likely to break things than to fix them. Why should this only be true for a bow? This will be executed for a bow regardless.

dktapps:
I think you ought to provide an explanation WHY this works.

PR #1593, opened 2016-08-03
Fix cannot shoot bow after jump

diamond-gold:
### Description
Fix cannot shoot bow after jump

### Reason to modify
It prevents $player->startAction from being set to -1 and data flag DATA_FLAG_ACTION from being set to false when server receive a PlayerActionPacket of type ACTION_JUMP when player jumps

### Tests & Reviews
Tested

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3025,8 +3025,10 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 						break;
 				}
 
-				$this->startAction = -1;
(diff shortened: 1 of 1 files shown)


dktapps:
Well, this seems better than #1587 . I'll test this.

dktapps:
Tested this, works flawlessly, and I can't find fault with your changes. Code Review +1.

ZedCee:
Tested.  +1

Oddly enough however, if you go into or out of fly mode, a primed bow will not work.

svilex:
Try sneaking while shooting the bow , same bug , also sneaking is an action ($player->startAction will be set to -1) . In singleplayer it is possible to shoot a bow after sneaking.

PR #1595, opened 2016-08-03
Change mcregion to anvil

KennanFatt:
### Description

Change mcregion to anvil for default format

### Tests & Reviews

anvil is default format level, right?


dschwartz783:
Actually... Can this ever equal null anyway? Doesn't it just use the default value if getProperty returns null? I think this entire if statement can be mostly refactored away.

ishitatsuyuki:
Code Review +1: Seems we missed that. Looks good.

dktapps:
+1 and merge

PR #1599, opened 2016-08-04
Fix updateAround bug

dog194:
### Description

Fix updateAround bug
updateAround bug
1. put a Jungle wood
2. put 4 cocoBean around this wood.
3. break the Jungle wood.
You will find that only one CocoBean update properly.
### Reason to modify

It should be.

### Tests & Reviews

Tested. But i donot will this cause other bug. 
can any one tell why this bug cause. or give an other solution. 

Fix updateAround bug


ishitatsuyuki:
No, you are probably breaking things.

robske110:
Probably not

ishitatsuyuki:
@robske110: ambiguous. Who are you responding to?

PR #1600, opened 2016-08-04
Fix furnace recipes.

dog194:
### Description

Any kind of wood can be smelt to charcoal now.

### Reason to modify

The furnace recipes are sent incorrectly.

### Tests & Reviews


dktapps:
If only NULL _did_ work the way it was supposed to. But this looks fine to me. Code Review +1.

ishitatsuyuki:
Code review +1

dktapps:
Tested this and can't find fault with it. Code Review +1.

ishitatsuyuki:
Code Review +1 again: thanks for removing the redundant glass one.

PR #1601, opened 2016-08-04
adjust spawn point to center of block

dschwartz783:
### Description

This PR changes world default spawn to the center of the block, instead of its corner, as per vanilla Minecraft.

### Reason to modify

Because a world we're working on would benefit from being able to accurately position the spawn point on a single block, rather than the middle of 4-blocks thing that's basically enforced now.

### Tests & Reviews

I have tested this, and I am now able to /setworldspawn to the middle of an individual block.

### DIFF
--- a/src/pocketmine/level/format/generic/BaseLevelProvider.php
+++ b/src/pocketmine/level/format/generic/BaseLevelProvider.php
@@ -97,7 +97,7 @@ abstract class BaseLevelProvider implements LevelProvider{
 	}
 
 	public function getSpawn(){
-		return new Vector3((float) $this->levelData["SpawnX"], (float) $this->levelData["SpawnY"], (float) $this->levelData["SpawnZ"]);
+		return new Vector3((float) $this->levelData["SpawnX"] + 0.5, (float) $this->levelData["SpawnY"], (float) $this->levelData["SpawnZ"] + 0.5);
 	}
 
 	public function setSpawn(Vector3 $pos){


dktapps:
Hmm. I think offset by 0.5x0.5 blocks would be better. I don't think this is a good idea. Spawning players to decimal coordinates can cause movement issues on player spawn.

dschwartz783:
Whoops, don't approve yet, one change needed.

dschwartz783:
@dktapps Except that some worlds might want the middle-of-4 blocks thing...

caspervanneck:
cool

caspervanneck:
i may test it

ZedCee:
Tested.  -1  
Doesn't seem to effect spawn point creation.  Still spawning in the middle of four blocks.

dschwartz783:
@ZedCee It works for me. Are you sure you tested correctly?

ZedCee:
@dschwartz783 Pardon my previous attempts.  I reset and repulled the branch and got it working.  See my new comment above.

ZedCee:
So it does partially work, but there is frequently a glitch of the player being dragged the the surface.  That seems like it may be problematic.

dschwartz783:
@ZedCee I haven't seen that. What coordinates are you using?

dktapps:
@dschwartz783 I think you should use ints, yes. Your way of doing things may also cause issues with the level.dat format. Use ints and compensate by adding +0.5 to the X and Z coordinates. That's how it's done in PC.

dktapps:
@caspervanneck This is a PR discussion, not a place for you to plead innocence or whatever.

dschwartz783:
I'm guessing that some PE world designs take the middle of 4 block spawn thing into account, which would be problematic for them. I really just don't like the inflexibility of hardcoding the spawn point orientation. This method may need some tweaking to work correctly in all circumstances, but let's do that then. Tell me where this fails and I or someone else can take care of it.

dktapps:
http://minecraft.gamepedia.com/Level_format
Spawn X, Y and Z are, or should be, ints.

ZedCee:
My coordinates come out with 2 decimal places, using `/setworldspawn' as player.
So for example something like `(278.9, 2, -359.46)`
If that still remains relevant.

dktapps:
https://github.com/iTXTech/Genisys/pull/1601/files#diff-3be932a70a48b8b5ec8042445dd0af86L100
I would say the only change you need to make is here. Add 0.5 to the X and Z coordinates when getting the world spawn. That way players will spawn in the center of the block instead of at its corner.

I also ought to observe that you don't need to allow doubles for the Y coordinate.

dschwartz783:
Why is it showing as two commits...?

dschwartz783:
You win. Simple adjustment :)

ZedCee:
I'm still finding the half block jump, albeit that could actually be  a little better from `master`.
This does however break spawn points between four blocks (good for symmetrical buildings).

Sorry @dschwartz783, I don't mean to be playing devil's advocate.

dschwartz783:
@ZedCee That's what I was saying earlier. Sigh...

dktapps:
Ah, well. Spawn positions in vanilla are in the center of the block, so this is the correct way to do it.

XFuryMCPE:
What if I want to set spawn inbetween 4 blocks?

dschwartz783:
@XFuryMCPE as @dktapps said, this is the best way to do it for the purposes of consistency with Minecraft clients, which do things this way. Could we offer an alternative? Sure, but this should be the default behavior.

XFuryMCPE:
I'm going to work on something. one minute

dktapps:
@XFuryMCPE you do that. But don't bother spamming another PR, and don't bother PRing until you **learn NOT TO USE GITHUB WEB**.

XFuryMCPE:
lol, I really do need to learn xD

Muqsit:
-1
I think the spawn must be set at exact coordinates. (Eg: x: 46.84327 y: 80.8421 z: 2.226)

dktapps:
@Muqsit Following the NBT vanilla format, that can't be done. The best that can be done is this. This is how it's done in vanilla.

gaoyichuan:
@Muqsit As in vanilla, the spawn should be at the center.

dktapps:
@gaoyichuan it was float to begin with, but I understand your point. Peculiar.

dschwartz783:
@ZedCee @Muqsit Please do not reject this. This is the correct way to do it. I understand your perspective, I shared it earlier, but @dktapps is correct in that this is how spawns are set in Minecraft, so this is how spawns should be set in Genisys.

Muqsit:
Okay, you're right. But Bukkit's /setworldspawn is accurate.

dschwartz783:
@Muqsit Bukkit isn't vanilla Minecraft either...

dschwartz783:
@all Okay, you know what? You guys talk amongst yourselves and decide how you want this done. I was working on precise spawn coordinates for setworldspawn originally, and it was rejected. Now I have vanilla compatibility, and it's being rejected too. Read the damn chat history before commenting. I think I understand why @PeratX force-merges so much...

ZedCee:
@dschwartz783 Relax.  "Read the damn chat history" doesn't really help your situation.  In my case, I was trying to gather information or type a response before 4 or 5 posts would blaze past me.  I also do not own MCPC, so I can't attest to it's functionality.  
@PeratX can force merge because of being the lead of the organization, but she shouldn't be doing that before things are properly tested and deliberated upon.  No one should.  That's how breakages happen.  PRs and branches are the most professional and code safe way to handle things in a collaborative coding environment.

  It's inline with MCPC and MCPE vanilla, but my OCD ain't happy about a lack of symmetry with my spawn point (my spawn point takes up exactly one chunk, 16x16).  I'm not the only one unhappy about this, think it's too bad it can't do both, and I feel any approval should be with a grain of salt.  

Tested.  The code works. +1

I understand I'm restating some of what was said and discussed upon, but I want to strongly stress the potential for fallout from players/users.

dktapps:
I meant to give this a +1 seems I never got to it

Muqsit:
I'm neutral about this.
What if someone does /setworldspawn 100.75 85 98.625
The spawn would be on 101.25 85 99.125.

dktapps:
@Muqsit you can't. Spawn coords can only be ints. This simply pads it to the center of the block.

ZedCee:
Here's a question; Does this affect any other form of spawn?  For instance re-logins...?

gaoyichuan:
Approved

PR #1603, opened 2016-08-04
Standardise $this->server calls

Striker209:
### Description

I was chilling around when I saw that

### Tests & Reviews

Now all Server calls are linked to $this->server instead of sometimes $server!

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -384,7 +384,7 @@ class Level implements ChunkManager, Metadatable{
 		$this->levelId = static::$levelIdCounter++;
 		$this->blockMetadata = new BlockMetadataStore($this);
 		$this->server = $server;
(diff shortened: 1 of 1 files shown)


dktapps:
Looks fine. +1

ishitatsuyuki:
Do you REALLY need this? PHP is a reference based language and `$this->server` and `$server` should be pointing to the same object.

ZedCee:
Was this even tested?  Could this break or fix anything?

ishitatsuyuki:
@ZedCee it doesn't break anything since it's just a local variable, but I think it also does nothing.

ZedCee:
That was the rhetorical question :wink:

Striker209:
I didn't know if I should have made pull request only with that modification but it's just for code clarity

ishitatsuyuki:
Closing: this is unneeded, and doesn't make it better or worse.

Issue #1594, opened 2016-08-03
addWindow / chests broken


robske110:
#### Issue description
I cannot open chests neither open inventories using the addWindow method

#### Steps to reproduce the issue
Try to open a chest->no inventory or whatsoever appears

#### OS and versions

* Genisys: https://github.com/iTXTech/Genisys/commit/45f2ae0840c270cb22c4e9af43abe550c4112593
* PHP: 7.0.6
* Server OS: MacOSX
* Game version: 0.15.4

#### Crashdump, backtrace, memory dumps, plugins list or other files

```
```

Muqsit:
Yeah, having the same issue since 0.15. We will investigate...

dktapps:
I sometimes get this on the Win10 branch. I haven't figured out what's causing it yet.

dktapps:
Noting that this happens _occasionally_ for me.

robske110:
Noting that this happens always for me on the master branch
Btw CS has no problems at all

gaoyichuan:
@robske110 Also CS has no feature at all

robske110:
Idc, i just pointed out that this is 100% Genisys's fault. We didn't change inventory for 0.15 in CS

dktapps:
@robske110 Nothing has changed inventory-wise whatsoever for 0.15 here either. I suggest you get your facts straight.

Muqsit:
Well, thanks for the hint @robske110 !

robske110:
Strange, I'll look at CS if we did anything inventory related in CS. And i'm just trying tonpoint out that this issue is a server side issue. If i find sth, I'll comment here.

Muqsit:
This issue has been fixed over here https://github.com/iTXTech/Genisys/pull/1605
Confirmed, and positive reviews too! (Lowkey want someone to +1 that pull).

dktapps:
@Muqsit I find it unlikely that that's the true solution.

dktapps:
This was fixed in https://github.com/iTXTech/Genisys/commit/1890dac91da327bd1bf5c5b86d41ae0b27a26328
Closing.

PR #1605, opened 2016-08-05
Fixes addWindow() not working properly #1594

Muqsit:
### Description
Fixes addWindow() not working most of the time.

### Reason to modify
I might sound dumb but I think int tags aren't working properly? Removing them literally fixed it...well at least for me!

### Tests & Reviews
After removing ints, addWindow() started to work flawlessly every single time. I am still *not sure* if this was the only reason behind addWindow() not working. I've tried adding a chest window and it worked.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -4327,11 +4327,10 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	 *
 	 * @return int
 	 */
-	public function addWindow(Inventory $inventory, int $forceId = null) : int{
+	public function addWindow(Inventory $inventory, $forceId = null){
 		if($this->windows->contains($inventory)){
 			return $this->windows[$inventory];
 		}
-
 		if($forceId === null){
 			$this->windowCnt = $cnt = max(2, ++$this->windowCnt % 99);
 		}else{
@@ -4343,7 +4342,6 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 			return $cnt;
 		}else{
 			$this->removeWindow($inventory);
-
 			return -1;
 		}
 	}


Muqsit:
Hmm let me test it with every tile.

dschwartz783:
@Muqsit If that's all it took to fix this, then I think we can safely say... #phpsucks

ishitatsuyuki:
@dschwartz783: no, if this really have effect, that's due to int being a built-in type (non-object), and thus cannot contain null.

dktapps:
Hmm. This shouldn't affect anything regardless because forceId is not used 99% of the time anyway. Peculiar.

Muqsit:
Yep. And confirmed. addWindow() worked 34 times in a row so I think it works every time now.

robske110:
Wait as far as i can see it should just be :float

robske110:
No lol

dktapps:
@robske110 what? WindowIDs are _always_ ints.

robske110:
Was confused by %99

robske110:
But why does this make it work again? I'll try it with float. If that works, we have a problem

Muqsit:
I don't get it too... but why do you think float is gonna work?

robske110:
Because apperently force convertinf the number to int breaks it. How about some var_dumping?

ishitatsuyuki:
@robske110: int probably rejects null.

robske110:
I'll just conduct some tests, ok? Will report what i found out.

dktapps:
I think we're running around in the dark here. I find it unlikely that this is the real cause or solution.

ishitatsuyuki:
Code Review +1.
If null is converted into int, it won't be compared correctly with === null.

ishitatsuyuki:
Oh, wait: the return value should be enforced to be int however, and please also correct the documentation.

robske110:
Yep, you got it. Due to forcing it to int and then being type strict (===) it fails. And this is why i don't like using this new php7 shit.

PR #1623, opened 2016-08-08
Fix ban command.

eddir:
### Description

Fix ban command bug.

### Reason to modify

Send the ban command with expire value:

> ban username short-reason 10

And you will receive an error:

 > Notice: Object of class DateTime could not be converted to int in phar:///root/19131/PocketMine-MP.phar/src/pocketmine/permission/BanEntry.php on line 74
[22:27:54] [Server thread/INFO]: An unknown error occurred while attempting to perform this command
[22:27:54] [Server thread/CRITICAL]: Unhandled exception executing command 'ban user reason 5' in ban: Call to a member function format() on integer
[22:27:54] [Server thread/CRITICAL]: Error: "Call to a member function format() on integer" (EXCEPTION) in "/src/pocketmine/permission/BanEntry" at line 93

### Tests & Reviews

Until variable must be instance of DateTime or null. Not int.

### DIFF
--- a/src/pocketmine/command/defaults/BanCommand.php
+++ b/src/pocketmine/command/defaults/BanCommand.php
@@ -53,7 +53,7 @@ class BanCommand extends VanillaCommand{
 		if(isset($args[0]) and isset($args[1])){
 			$reason = $args[0];
 			if($args[1] != null and is_numeric($args[1])){
-				$until = $args[1] * 86400 + time();
+				$until = new \DateTime('@' . ($args[1] * 86400 + time()));
 			}else{
 				$until = null;
 			}
--- a/src/pocketmine/command/defaults/BanCommand.php
+++ b/src/pocketmine/command/defaults/BanCommand.php
@@ -59,8 +59,9 @@ class BanCommand extends VanillaCommand{
 			}
 
 			$sender->getServer()->getNameBans()->addBan($name, $reason, $until, $sender->getName());
+		}else{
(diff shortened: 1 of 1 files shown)


dktapps:
Interesting. I never even knew this crashed until now. I'll test this.

dktapps:
It fixes the crash. Can't say whether it fixes the actual time banning or not. Have you tested that?

eddir:
@dktapps You are right. There is a new bug :D

dschwartz783:
Keep in mind that this bans in increments of a day, so "/ban username reason 1" means, banned for one day. I have tested with "/ban username reason 0" and it kicks/bans, but since the ban immediately expires, it lets me back in. I also removed the " * 86400" part, and tried to ban myself for 10 seconds. Seems to be working correctly. I was unable to join for 10 seconds, after which it expired.

dschwartz783:
Although, he committed on GitHub, so he might not be able to squash...

ZedCee:
Tested.  Can confirm with @dschwartz783.  Also accepts up to 2 decimal places without a crash, mind you the math doesn't seem right when you do that.

ishitatsuyuki:
Seems this looks OK. Any point to review?

dktapps:
This fixes a critical error, so this seems fine. It's not his fault that the time banning is a total load of #### to begin with. I'll rewrite it at a later date. +1

ishitatsuyuki:
+1

PR #1624, opened 2016-08-08
implement PacketReliability constants from RakNet protocol

dschwartz783:
### Description
This should enhance RakLib's readability a bit. No more mysterious reliability literals floating around the codebase... I think I got everything down and coded decently enough... This was all taken from the RakNet protocol.

### Reason to modify
Because life is too short to waste it spending time figuring out what a literal is for.

### Tests & Reviews
I've run some tests, and this doesn't seem to break anything. Tell me if I'm wrong though. :)

### DIFF
--- a/src/pocketmine/network/RakLibInterface.php
+++ b/src/pocketmine/network/RakLibInterface.php
@@ -29,6 +29,7 @@ use pocketmine\Player;
 use pocketmine\Server;
 use pocketmine\utils\MainLogger;
 use raklib\protocol\EncapsulatedPacket;
+use raklib\protocol\PacketReliability;
 use raklib\RakLib;
 use raklib\server\RakLibServer;
 use raklib\server\ServerHandler;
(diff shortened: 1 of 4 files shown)


dktapps:
I think you'd be better to use `self::` instead of `$this::` since you're accessing constants.
Please correct this. Rejecting until corrected.

dschwartz783:
Okay, think I wacked all those moles. :)

dktapps:
Yep, that looks fine. Please squash once you're done.

dschwartz783:
Did

dktapps:
Ah, browser didn't reload properly. Code Review +1.

ishitatsuyuki:
I think using inheritance is a bad idea. Accessing as static constants in a separate class will be better.

dschwartz783:
I figured I'd use the same method used by things like the BlockId interface and ItemId interface.

dktapps:
@dschwartz783 The reason I used interfaces for blockIds/itemIds was because they needed to be accessible with `self::` from those classes for backwards compat reasons. Here, that doesn't matter. So you can just import the class/interface and use it with `PacketReliability::RELIABILITY_WHATEVER`

dktapps:
:+1:

dschwartz783:
Okay, everything's swapped out, inheritance is no longer used.

dktapps:
+1

ishitatsuyuki:
Please use class since interface is just an ugly feature of those language without multiple inheritance.

dschwartz783:
@ishitatsuyuki Will do. If only PHP had structs... smh...

ishitatsuyuki:
Code review +1

dktapps:
+1 and merge.

PR #1630, opened 2016-08-10
Fix #1583

### LINKED ISSUE #1583: Messages containing %0 can freeze/crash the client
#### Issue description
If a player do /me %0 , it will crash all connecteds clients. Bug tested on many servers.

1' Do /me %0

#### OS and versions

* Genisys: 6b404b2
* PHP: 7.0.3
* Server OS: On my Android phone
* Game version: 0.15.4

```
```

dktapps:
### Description
Fixed #1583 -  TextPackets with type TYPE_TRANSLATION and containing %0 causes clients to freeze/crash.

### Reason to modify
Remove exploits used to crash server players.

Some translations are handled automatically by the client if allowed to. My belief is that sending translation key %0 causes an eternal loop while the client searches for a translation string that doesn't exist (#blamemojang). This fixes that by removing %0 from any translation keys.

### Tests & Reviews
I have tested the code and it works. Tested both as client and console, and this can no longer be exploited.

### DIFF
--- a/src/pocketmine/lang/BaseLang.php
+++ b/src/pocketmine/lang/BaseLang.php
@@ -92,7 +92,7 @@ class BaseLang{
 			$baseText = str_replace("{%$i}", $this->parseTranslation((string) $p), $baseText, $onlyPrefix);
 		}
 
-		return $baseText;
+		return str_replace("%0", "", $baseText); //fixes a client bug where %0 in translation will cause freeze
 	}
 
 	public function translate(TextContainer $c){


TheAppleGamer:
Works just fine for me

FaanMaario:
Works fine

ishitatsuyuki:
Looks OK. Code Review +1

ZedCee:
Tested.  Doesn't appear to break anything and prevents a pretty volatile exploit. +1

PR #1646, opened 2016-08-14
fix for Economys

caspervanneck:
### Description
 economys plugin dont work we get en eror 

### Reason to modify
we get a eror and i try to fix it

### Tests & Reviews
 plugin works fine now 

 it works good

 tish is the eror  (pocketmine\utils\LevelException: "Undefined Level reference" (EXCEPTION) in "/src/pocketmine/level/Position" at line 114)  i fix it (:

### DIFF
--- a/src/pocketmine/level/Position.php
+++ b/src/pocketmine/level/Position.php
@@ -76,30 +76,6 @@ class Position extends Vector3{
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


dktapps:
assert() in production environment? -1

ZedCee:
Committed on GitHub? -1
*How did you test this?*

PR #1647, opened 2016-08-14
Update for Universal 0.15

Adam1609:
### Description
 Update to all 0.15 version

### Reason to modify
 Because 0.15.6 could join also.

### Tests & Reviews

 It works now!


dktapps:
This is completely unnecessary. **Have you made sure that there is actually a problem before trying to fix it?**

Adam1609:
No problem!

gaoyichuan:
Actually I want to ask: can you understand html?

PR #1648, opened 2016-08-14
Add composer

mhsjlw:
### Description
Adding Composer to Genisys for managing releases and packages

### Reason to modify
[NIH is a big deal](https://en.wikipedia.org/wiki/Not_invented_here) seriously harming PHP developers all around the globe. Fortunately for you, there is a fix, [Composer](https://getcomposer.org/). This pull request adds Composer to Genisys allowing for package management and just a little more sanity to the project. However, this fix comes at a cost. After adding *any* dependencies, each user who will use Genisys from now on will have to have Composer installed on their system to run Genisys and/or build it to a phar, this does not include pre-built phars, any phars built with Composer will bundle up all the dependencies.

This also opens up some awesome feature capabilities that may have not been previously available without a rewrite from scratch:

- [php-a-star](https://github.com/jmgq/php-a-star) An implementation of the famous A* search algorithm in PHP. This allows you to easily and quickly build mob AI for following players and an AI for animals that includes running away from an attacker
- [php-jwt](https://github.com/firebase/php-jwt) A JWT implementation in pure PHP (the token used in logging in) also includes verifying signatures from logging in players
- [phpecc](https://github.com/phpecc/phpecc) A PHP library that adds support for a bunch of elliptic curves, used in Minecraft: PE's encryption. This will allow a quick and painless implementation of MC: PE encryption in Genisys

... *and the lines become more blurred*

### DIFF
--- a/tests/ci.sh
+++ b/tests/ci.sh
@@ -1,4 +1,9 @@
 #!/bin/bash
+php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+php -r "if (hash_file('SHA384', 'composer-setup.php') === 'e115a8dc7871f15d853148a7fbac7da27d6c0030b848d9b3dc09e2a0388afed865e6a3d6b3c0fad45c48e2b5fc1196ae') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
+php composer-setup.php
+php -r "unlink('composer-setup.php');"
+./composer.phar install
 mkdir plugins
 curl -fsSL https://github.com/iTXTech/Genisys-DevTools/releases/download/1.0.0/Genisys-DevTools_v1.0.0.phar -o plugins/Genisys-DevTools.phar
 echo Running lint...
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ Desktop.ini
 /docs/build/
 !/docs/requirements.txt
 .project
+
+/vendor/
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ php:
  - 7.0
 
 before_script:
+ - php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+ - php -r "if (hash_file('SHA384', 'composer-setup.php') === 'e115a8dc7871f15d853148a7fbac7da27d6c0030b848d9b3dc09e2a0388afed865e6a3d6b3c0fad45c48e2b5fc1196ae') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
+ - php composer-setup.php
+ - php -r "unlink('composer-setup.php');"
+ - ./composer.phar install
  - phpenv config-rm xdebug.ini
  - pecl install channel://pecl.php.net/pthreads-3.1.6
  - pecl install channel://pecl.php.net/weakref-0.3.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,6 @@ php:
  - 7.0
 
 before_script:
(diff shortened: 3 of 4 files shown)


ishitatsuyuki:
We don't have enough reason to use that. OpenSSL still looks suitable, please add more reason if you can.

tests/ci.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash
+php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+php -r "if (hash_file('SHA384', 'composer-setup.php') === 'e115a8dc7871f15d853148a7fbac7da27d6c0030b848d9b3dc09e2a0388afed865e6a3d6b3c0fad45c48e2b5fc1196ae') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
+php composer-setup.php
+php -r "unlink('composer-setup.php');"
+./composer.phar install
```
ishitatsuyuki: We will prefer a systemwide install in Docker.

mhsjlw:
Sorry, what? AI, Encryption and JWT signing isn't *enough*, along with all the other 106, 559 possibilities for packages? This only effects a Genisys developer's experience not the end user's experience. I really don't understand how an extra step isn't worth not re-inventing the wheel.

ishitatsuyuki:
Some doubts:
* JWT can be implemented with string splitting and OpenSSL for verifying.
* So is encryption.
* And... I need something other than A*.

ZedCee:
Could you elaborate on your quarrels with A*, @ishitatsuyuki?  A fresh AI would be really nice, or for better or worse at least being able to enable all the mobs.

ishitatsuyuki:
Things regarding the AI:
* Who will write them? (I currently don't have plan)
* Is A* the most suitable algorithm?
* Why not even do a custom implementation by ourself?

ZedCee:
Just my own personal opinions:

- Exactly.  And I can't code them, I wouldn't know where to begin.  We don't have a solid plan past what seems like waiting for CS.  We need one.
- It's possible there are others; could composer still be viable?  If we do implemented composer, A* could still be viable as a placeholder AI until we get things going.
- Again, this comes down to the plan.  What's the plan?

ishitatsuyuki:
Closing since we still don't need it and the way to install it in CI is wrong. We will add it when we need it.

gaoyichuan:
Actually I love composer. Since by using composer we won't have to manage the php extensions by our own code.

dktapps:
@ZedCee Waiting for CS? What? Are you mental?

ZedCee:
Well I did get the reaction I was expecting...  So I suppose I'm not mental.
I do not endorse CS, nor do I use it. (Don't really follow them either)
But they have been in our forums spouting about their AI they're gonna be working on.
We need a plan.

mhsjlw:
Here's the thing. AI (believe it or not) isn't a piece of cake and certainly *reinventing an entire library* doesn't make things any easier. But if you want to get this done, which as of right now, no one seems to care at all about, if you use the A* library, I promise you that your AI will be faster, easier to understand and more stable. Although, what do I know? I'm just someone with experience writing 2D pathfinding AI (but not in PHP, thankfully). @ishitatsuyuki isn't interested in merging it becuase he's convinced he's got everything figured out so you might as well leave this closed and lock the conversation becuase I'm not wasting any more of my time

ishitatsuyuki:
I'm just closing this because we don't need this for now and this doesn't benefit anyone other than the team. Additionally, your patch wasn't meeting our way to use things in CI, thus I don't want to merge this PR. Things will just get added, when someone uses it.

mhsjlw:
Sounds good by me. So you'd OK a PR adding encryption that requires composer

ishitatsuyuki:
Well, if you can, I don't bother that, but OpenSSL is enough for encryption...

PR #1656, opened 2016-08-15
Implement hoppers (pickup items, pull items, feed items into target inventories)

dktapps:
## Description
Added hoppers, with the ability to pull items from tile inventories, pick up dropped items and feed items into inventories.

Hoppers will pull items down from any *permanent* tile inventory above them, and also pick up dropped items. Directional feeding also works - i.e. hoppers will feed into the inventory they are pointing at.

### Known issues
- Powering a hopper does not currently prevent it processing items. This is because of the shoddy implementation of redstone making it difficult to activate/deactivate blocks.
- All slots of inventories are currently treated the same. This means that hoppers can, for example, feed into furnace result slots and pull items out of the target and/or fuel slot. This will be corrected at a later date when I upgrade inventory and crafting base functionality.
- Hopper locking (only opens when player clicks on the hopper with an item with a matching custom name) is impossible to use because of the lack of a /blockdata command. The functionality has however been implemented.

## Reason to modify
People want hoppers, so let's _give_ them hoppers.

## Tests & Reviews
I have done my best to ensure that the behaviour follows the behaviour defined [here](http://minecraft.gamepedia.com/Hopper) ~~with a few exceptions to line up with current MCPE hopper behaviour.~~ EDIT: I've had information from certain sources which leads me to believe that current MCPE hopper pickup behaviour is bugged, so I have followed the PC guidelines instead.
To the best of my knowledge, everything works fine. Tested on Windows 10, with Win10 Edition and MCPE. Please test and confirm.

### DIFF
--- a/src/pocketmine/tile/Hopper.php
+++ b/src/pocketmine/tile/Hopper.php
@@ -1,0 +1,309 @@
+<?php
+
+/*
+ *
+ *  _____   _____   __   _   _   _____  __    __  _____
+ * /  ___| | ____| |  \ | | | | /  ___/ \ \  / / /  ___/
+ * | |     | |__   |   \| | | | | |___   \ \/ /  | |___
+ * | |  _  |  __|  | |\   | | | \___  \   \  /   \___  \
+ * | |_| | | |___  | | \  | | |  ___| |   / /     ___| |
+ * \_____/ |_____| |_|  \_| |_| /_____/  /_/     /_____/
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author iTX Technologies
+ * @link https://itxtech.org
+ *
+ */
+
+namespace pocketmine\tile;
+
+use pocketmine\block\Hopper as HopperBlock;
+use pocketmine\entity\Item as DroppedItem;
+use pocketmine\inventory\HopperInventory;
+use pocketmine\inventory\InventoryHolder;
+use pocketmine\item\Item;
+use pocketmine\level\format\FullChunk;
+use pocketmine\math\Vector3;
+use pocketmine\nbt\NBT;
+use pocketmine\nbt\tag\CompoundTag;
+use pocketmine\nbt\tag\IntTag;
+use pocketmine\nbt\tag\ListTag;
+use pocketmine\nbt\tag\StringTag;
+
+class Hopper extends Spawnable implements InventoryHolder, Container, Nameable{
+	/** @var HopperInventory */
+	protected $inventory;
+
+	/** @var bool */
+	protected $isLocked = false;
+	
+	/** @var bool */
+	protected $isPowered = false;
+
+	public function __construct(FullChunk $chunk, CompoundTag $nbt){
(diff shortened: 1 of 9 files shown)


XFuryMCPE:
Lol. "People want hoppers, so let's give them hoppers."

Muqsit:
Nice. :)

ZedCee:
Just wrapping up some of the finishing touches folks.  
I'm testing now and have been using them throughout development.
So far things seem decent, forgive some bugs we just have to squash.

dktapps:
You mean like server spammed to its limits when you blow up a hopper? Yea, that does get inconvenient.

Muqsit:
It causes a lot of lag :3

dktapps:
@Muqsit How big a server are you running? I've had no issues while testing. I'll add configuration options in if you want.

Muqsit:
@dktapps 60 players (mcregion, max memory: 6000 gb, server uses 4 cores).
Players usually get stuck on the "Building Terrain" screen.

dktapps:
@Muqsit Any terminal error? __clone on null or whatever, I experienced similar issues, fixed them since.

src/pocketmine/tile/Hopper.php
```diff
@@ -0,0 +1,309 @@
...
+				$inventory->removeItem($item);
+				$this->resetCooldownTicks();
+				if($source instanceof Hopper){
+					$source->resetCooldownTicks();
+				}
+			}
+		}
+		
+		//Feed item into target inventory
+		//Do not do this if there's a hopper underneath this hopper, to follow vanilla behaviour
```
ishitatsuyuki: Wait, this is not pure vanilla behavior IIRC, will this cause minor issues?

dktapps: Not _pure_, but it was the only simple way I could get it to behave the way vanilla hoppers do.

Muqsit:
@dktapps I'll hard test hoppers. So far, the console is clean.

src/pocketmine/tile/Hopper.php
ishitatsuyuki: Not only below but it should be based on property I think, hopper has output direction.

dktapps: Yes, it does. That's already implemented. It just skips feeding in this circumstance.

ZedCee:
Tested throughout the development.  The rebase of the w10 fixes seems to play quite nice, with no apparent duplication. +1

#### OS and versions
* Genisys: 
https://github.com/iTXTech/Genisys/pull/1656/commits/22f2e28f544bee34dd9d1e2f85b36fd3f3fba844
* PHP: 7.0.9 (from Arch)
* Server OS: Arch Linux
* Game version: 0.15.6 iOS and Android

ishitatsuyuki:
Cool, seems you are right: hoppers instantly feed items into the target inventory if stacked vertically but not horizontally.

ishitatsuyuki:
+1

Muqsit:
@dktapps I tried to overload my test server, but the "Building Terrain" thingy still didn't happen. Probably a fault with my server.

dktapps:
@Muqsit As long as the terminal was clean. If anything dangerous crashed you'd know about it. If anyone else encounters similar issues, I'm sure we'll find out pretty quickly :D

PR #1668, opened 2016-08-17
Added Effect Events (EntityEffectAddEvent, EntityEffectRemoveEvent)

dschwartz783:
### Description
This adds the ability for plugins to listen to EntityEffectAddEvent and EntityEffectRemoveEvent, which was requested by some guy on Gitter. Figured I'd throw it in here, since it may come in handy. :)

### Reason to modify
These events don't exist at the moment, and may come in handy, so here they are!

### Tests & Reviews
I've tested these events, and they work as advertised. The command is also silent when the events are cancelled, and works correctly otherwise.

### DIFF
--- a/src/pocketmine/command/defaults/EffectCommand.php
+++ b/src/pocketmine/command/defaults/EffectCommand.php
@@ -114,13 +114,15 @@ class EffectCommand extends VanillaCommand{
 				return true;
 			}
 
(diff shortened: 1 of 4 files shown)


ishitatsuyuki:
Hmm, looks good. Only one point: please use capitalized commit title, probably next time.
+1

Muqsit:
+1

PR #1676, opened 2016-08-17
Painting: Fix rotation when placed.

### LINKED ISSUE #1138: Paintings rotate when placed + wrong positioning
The search bar doesn't return any result for me with keyword "Painting" and there aren't any Painting issue open in the Issues section.

I just wonder if your Painting rotate work well? It ALWAYS rotate to be in the wall. See the attachment.

untitled

dktapps:
### Description
Fixed paintings rotating when placed. Close #1138

### Reason to modify
Entity movement is updated when creating the entity. This should not happen for paintings since paintings cannot move. This PR corrects that problem by overriding the method with an empty method.

### Tests & Reviews
Tested, found fully working with PE 0.15.6.

### DIFF
--- a/src/pocketmine/entity/Painting.php
+++ b/src/pocketmine/entity/Painting.php
@@ -45,6 +45,10 @@ class Painting extends Hanging{
 		parent::spawnTo($player);
 	}
 
+	protected function updateMovement(){
+		//Nothing to update, paintings cannot move.
+	}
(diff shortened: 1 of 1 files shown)


ZedCee:
Tested. +1

ishitatsuyuki:
Code Review +1

PR #1678, opened 2016-08-17
Added PlayerJumpEvent

Tarik02:
### Description

Added PlayerJumpEvent, useful for plugins.

### Reason to modify

I'm want to use this event in my anti-cheat and other plugins.

### Tests & Reviews

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -70,6 +70,7 @@ use pocketmine\event\player\PlayerHungerChangeEvent;
 use pocketmine\event\player\PlayerInteractEvent;
 use pocketmine\event\player\PlayerItemConsumeEvent;
 use pocketmine\event\player\PlayerJoinEvent;
+use pocketmine\event\player\PlayerJumpEvent;
 use pocketmine\event\player\PlayerKickEvent;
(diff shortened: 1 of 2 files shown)


XFuryMCPE:
There's no way to use it, you didn't even add any functions to the event.

dktapps:
@XFuryMCPE How about getting your facts straight before making derogatory comments about others' work? I might add that at least he took the time to learn GIT instead of complaining and asking others to do his work for him. 

This would act as a notification for plugins.

gaoyichuan:
This should waste a lot of resources.

ishitatsuyuki:
I don't think this waste resources, but however it's incomplete. No cancel function is provided.

dktapps:
@ishitatsuyuki you wouldn't be able to cancel it anyway. That's client handled.

MrGenga:
@dktapps no you can, just setMotion (0, -$jumpHigh, 0)

XFuryMCPE:
@dktapps I don't see why you're so picky about whether git is used or not. It's not the end of the world if git isn't used, 'oh no, there are 2 commits for one feature', who cares??? Half the people downloading this don't even use GitHub or look at the commits.

PR #1684, opened 2016-08-18
You must support PHP without openssl!

fengberd:
### Description
To support PHP without openssl.

### Reason to modify
Too many users is using PHP binaries without openssl,this will available them to continue use this core.


ishitatsuyuki:
No, we won't support. We have provided updates for any platform, don't get stick with ancient things.

PR #1686, opened 2016-08-18
Level::dropItem now returns the dropped item

Tarik02:
### Description

Now method Level::dropItem returns dropped item.

### Reason to modify

I'm want to store dropped items, this allows o do this.

### Tests & Reviews


ishitatsuyuki:
Good idea. +1

dktapps:
+1 and merge.

PR #1687, opened 2016-08-18
Added reason to PlayerQuitEvent

Tarik02:
### Description

This PR adds reason, why player leave the server(client disconnect, timeout(maybe crash)).

### Reason to modify

I'm want to get reason, how player left the server.

### Tests & Reviews


dktapps:
This is redundant. The quit message will tell you this anyway.

Tarik02:
@dktapps, quit message is always same and not says the reason.

PR #1700, opened 2016-08-19
Change time query cmd and list perm to true

ABAxStorm:
### Description

This PR sets the permissions for the commands /time query and /list to TRUE

### Reason to modify

Allowing players to use '/time query' causes no problems (could be useful for survival), and /list could be helpful as well for players

### Tests & Reviews

This follows the correct syntax


dog194:
I think it's no a good idea that change the permissions for these commands.
Mmmmmm... I think if some server owner has some special demand like these, they should use plugins.

ishitatsuyuki:
This is not the vanilla behavior. Closing.

PR #1703, opened 2016-08-19
Improve random bytes generation.

minecrafter:
### Description

This PR changes all usages of `Utils::getRandomBytes()` to the builtin function `random_bytes()`. `Utils::getRandomBytes()` has been kept, but is deprecated and simply wraps `random_bytes()`.

### Reason to modify

* The existing function is poor-quality, mixing in only bad sources of entropy.
* Most of the "secure" random sources used read from the same ultimate source (/dev/urandom on Linux, macOS, and the BSDs), and the "weak" sources are extremely poor-quality.
* PHP 7 has a function to return cryptographically-secure random bytes, so the mess of code is not needed since we're using PHP 7.
* No calls are made for secure random bytes, ever.

### Tests & Reviews


ishitatsuyuki:
Good, please squash the commits.

minecrafter:
Commits squashed as requested.

ishitatsuyuki:
Code review +1

gaoyichuan:
Code review +1 and merge to master

MrGenga:
It not need to OpenSSL? YAY!

PR #1708, opened 2016-08-19
Use Online API if there is no OpenSSL extension

MrGenga:
### Description

Very useful thing, if you haven't updated php binaries, and you want XBOX auth, and if you have good internet connection, you can use that!

### Reason to modify

I think it is the best way, because it is crossplatform. If I used /usr/bi /openssl, there is no Windows support.
### Tests & Reviews

I think to move API to more stable server, then I create new pull request, with url change.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -1892,8 +1892,8 @@ class Server{
 			]);
 
 			$onlineMode = $this->getConfigBoolean("online-mode", false);
-			if(!extension_loaded("openssl")){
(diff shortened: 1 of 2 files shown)


minecrafter:
This sounds extremely scary, sending encrypted data and getting it back over an unencrypted connection, plus it's a remote server that I can't trust.

I would not recommend merging this PR.

minecrafter:
#1703 is a much better solution (using the native PHP 7 `random_bytes()` function).

gaoyichuan:
DO NOT use spaces, use TABs for your lines

ZedCee:
@gaoyichuan That comes from using GitHub editor, hence why it is so strongly recommended against. ;)

PR #1710, opened 2016-08-20
OpenSSL is not need now! Utils::getRandomBytes has been replaced with…

MrGenga:
### Description

This PR removed openssl checking, because it isn't required for now, because Utils::getRandomBytes has been replaced with random_bytes

### Reason to modify

Only delete openssl checking.
### Tests & Reviews

There is no tests, because it only delete one if(){}

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -1892,10 +1892,7 @@ class Server{
 			]);
 
 			$onlineMode = $this->getConfigBoolean("online-mode", false);
-			if(!extension_loaded("openssl")){
-				$this->logger->warning("The OpenSSL extension is not loaded, and this is required for XBOX authentication to work. If you want to use Xbox Live auth, please update your PHP binaries at itxtech.org/download, or recompile PHP with the OpenSSL extension.");
-				$this->setConfigBool("online-mode", false);
-			}elseif(!$onlineMode){
+if(!$onlineMode){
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
You are idiot. At least do a grep before doing such a thing.

ishitatsuyuki:
Test your code, even for a trivial change. You're just breaking things.

MrGenga:
@ishitatsuyuki I tested it, and it works. I can join with XBOX, but when I logout I get message about authenticate on XBOX Live. Test server is running for now: gc-game.ru:12228
P. S. I doesn't have OpenSSL in my binary, but XBOX Live authenticate working well!

ishitatsuyuki:
No, without the binary it will emit warning or silently fail. You either has OpenSSL built in or disabled online-mode.

MrGenga:
@ishitatsuyuki online-mode true, in older Genisys I had error about unknown function openssl_random_pseudo_bytes.

ishitatsuyuki:
Anyway, this change will definitely not work. I don't want to spend time anymore, and you should check your environment. Check if your CURRENT PHP includes OpenSSL.

minecrafter:
OpenSSL is still needed for Xbox authentication (I did not "remove" the dependency in #1703), and the `random_bytes` replacement affected only limited parts of the code base (not enough to seriously matter).

PR #1712, opened 2016-08-20
The $immediate parameter was added in Player

legoboy0215:
### Description

The $immediate parameter was added in Player, but not modified in this class.

### Reason to modify

This causes a warning since synapse\Player extends pocketmine\Player, and the parameters should match.

### Tests & Reviews
Yes, I have tested this and this fixes the warnings.

https://github.com/iTXTech/Genisys/blob/master/src/pocketmine/Player.php#L1152

### DIFF
--- a/src/synapse/Player.php
+++ b/src/synapse/Player.php
@@ -270,8 +270,8 @@ class Player extends PMPlayer{
 		$this->uuid = $uuid;
 	}
 
-	public function dataPacket(DataPacket $packet, $needACK = false){
-		$this->interface->putPacket($this, $packet, $needACK);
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
Please also modify the directDataPacket, to follow DRY.

PeratX:
No need. directDataPacket is immediate.

PeratX:
I will push a fix later.

PeratX:
Wait, who has modified this method???

legoboy0215:
So did I not follow the format or what...

PeratX:
I have pushed a fix

legoboy0215:
Ok.

PR #1713, opened 2016-08-20
Fix russian translation with XBOX

MrGenga:
### Description

Add new translations to russian translation files

### Reason to modify

Because, I am russian, I can do it :3
### Tests & Reviews

There is no code to test, because it is not code ;D


daniktheboss:
Молодец

ishitatsuyuki:
Seems good, let's merge it.

PR #1715, opened 2016-08-20
Re-add InventoryTransactionEvent with slightly different implementation

dktapps:
### Description
New implementation of InventoryTransactionEvent. Fix plugin crashes.

**Please note: the reason this was removed in the first place was because it's inaccurate. With Win10 in the mix, slot changes cannot be paired.** This event will now be called when the player's transaction queue starts being processed. Cancelling this event will prevent queue execution.

### Reason to modify
Fix plugin crashes and provide a measure of control for plugins over the new inventory transaction system. This is not an ideal way to do it, but it's the best that can be done for now.

### Tests & Reviews
I have tested functionality and found no faults. However, I have not tested this with a plugin that uses the old-style event. Please test and review.

### DIFF
--- a/src/pocketmine/event/inventory/InventoryTransactionEvent.php
+++ b/src/pocketmine/event/inventory/InventoryTransactionEvent.php
@@ -1,0 +1,60 @@
+<?php
+
+/**
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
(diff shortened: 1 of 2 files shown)


PeratX:
+1

Muqsit:
+1
P.S: Thanks! =D

PR #1717, opened 2016-08-20
Fix my OMG error

MrGenga:
### Description

Fix FastPlayerListPacket.

### Reason to modify

ERROR FIX!
### Tests & Reviews

Only fix.


dktapps:
Fail. No files changed?

MrGenga:
@dktapps you already fix it, lol :3

dktapps:
I might also add that that error would never have occurred in the first place if your code had been properly tested before PR or merge. I'm not impressed.

PR #1720, opened 2016-08-20
Fix crashs with FastPlayerList update

MrGenga:
### Description

It fix crash with error about FastPlayerListPacket

### Reason to modify

Fix crash :)
### Tests & Reviews

I have tested the code and it works. Client doesn't freezes I check it, transfer per 4 servers and it doesn't freeze.

### DIFF
--- a/src/synapse/Player.php
+++ b/src/synapse/Player.php
@@ -278,8 +278,10 @@ class Player extends PMPlayer{
 			$pk->sendTo = $this->uuid;
 			$pk->type = $packet->type;
 			foreach($packet->entries as $entry){
-				array_pop($entry);
-				array_pop($entry);
+				if(!$packet->type == PlayerListPacket::TYPE_REMOVE){
+					array_pop($entry);
+					array_pop($entry);
+				}
 				$pk->entries[] = $entry;
 			}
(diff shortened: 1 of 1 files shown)


dktapps:
DETAILS. What errors were caused, and why do your changes solve the problem? Also please squash your commits.

MrGenga:
@dktapps I will ad note to code lines, ok

src/synapse/Player.php
```diff
@@ -278,8 +278,10 @@ public function dataPacket(DataPacket $packet, $needACK = false){
 			$pk->sendTo = $this->uuid;
 			$pk->type = $packet->type;
 			foreach($packet->entries as $entry){
-				array_pop($entry);
-				array_pop($entry);
+				if($packet->type !== PlayerListPacket::TYPE_REMOVE){
```
MrGenga: There is error, because entry has 2 unwanted parametrs, but only in TYPE_ADD, I added check of type to fix it.

PeratX:
He fixed different types of PlayerListPacket.

PeratX:
+1

dktapps:
+1 and merge.

PR #1721, opened 2016-08-20
fix bug ,resultitem will be droped when a creative player with full i…

Alcatraz-Du:
### Tests & Reviews

…nventory crafts items


dktapps:
Seems okay, but please provide the required details in future. I'll let you off this time since it's a very small change. +1.

ishitatsuyuki:
Isn't this the supposed behavior?

dktapps:
@ishitatsuyuki no, try it in vanilla MCPE. It drops the excess regardless of gamemode.

ishitatsuyuki:
Approved.

PR #1727, opened 2016-08-21
Improve deprecated methods

MrGenga:
### Description

Some old plugins uses setHotbarSlotIndex or setHeldItemSlot.

### Reason to modify

To work old plugins.
### Tests & Reviews

After W10 update, first 9 inventory slots automaticly at hotbar. If plugin want to change it, item in $slot moved to hotbar slot, and item in that slot has been moved to $slot. Same in setHeldItemSlot.

### DIFF
--- a/src/pocketmine/inventory/PlayerInventory.php
+++ b/src/pocketmine/inventory/PlayerInventory.php
@@ -99,7 +99,10 @@ class PlayerInventory extends BaseInventory{
 	 * Changes the linkage of the specified hotbar slot. This should never be done unless it is requested by the client.
 	 */
 	public function setHotbarSlotIndex($index, $slot){
-		trigger_error("Do not attempt to change hotbar links in plugins!", E_USER_DEPRECATED);
+		$i1 = $this->getItem($index);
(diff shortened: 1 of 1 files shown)


ZedCee:
So this has this been tested on W10?  setHotbarSlotIndex or setHeldItemSlot we're depreciated to extend support...  
I've tested on PE, doesn't appear to break anything, seems to at least partially work for some plugins.  Albeit, I'm not sure the best ones to test this with.  
How have you gone about testing this?

MrGenga:
@ZeedCee it doesn't break a things, because it DOESN'T REMAP HOTBAR SLOTS, only move items to it

ishitatsuyuki:
You should not remove the deprecate warning.

MrGenga:
@ishitatsuyuki @deprecated?

MrGenga:
@dktapps why? It is adapted for W10, look at the code

nyankichi820:
@MrGenga ++

Although the order of win10 support , I think it is not good to deprecate without generally available means .

PR #1728, opened 2016-08-21
Fix the case of characters ( XBOX -> Xbox )

nao20010128nao:
### Description
Fix the case of characters. And removed "Live" from Russian lang file.

### Reason to modify
- See [xbox.com](xbox.com) for correct case of "Xbox".
- This PR is to follow the correct case on [xbox.com](xbox.com).


ishitatsuyuki:
Hmm, I followed logo for the case...

ZedCee:
Approve +1

nao20010128nao:
Thanks

PR #1739, opened 2016-08-22
fix traditional Chinese

edwinyoo44:


ishitatsuyuki:
Okay, looks good. Merging.

PR #1751, opened 2016-08-22
Fix #1744 - disallow teleporting sleeping players

### LINKED ISSUE #1744: TP + Bed :'D
#### Issue description
TP + Bed

1. sleep bed
2. tp

#### OS and versions
* Genisys: https://github.com/iTXTech/Genisys/tree/3d8f4377f30907ca8b01c2de6fcb7a7a477941c6
* PHP: 7.0.10 (Genisys script)
* Server OS: Ubuntu 14.04.4 LTS
* Game version: 0.15.6 (Android)

14114084_837320386398495_436433307_o

dktapps:
### Description
Disallow teleporting sleeping players. Fix #1744 

### Reason to modify
Fix #1744 

### Tests & Reviews
This has not been tested, but it should fix the problem. Please test and review.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -4274,7 +4274,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	 * @return bool
 	 */
 	public function teleport(Vector3 $pos, $yaw = null, $pitch = null){
-		if(!$this->isOnline()){
+		if(!$this->isOnline() or $this->isSleeping()){
 			return false;
 		}
 
--- a/src/pocketmine/command/defaults/TeleportCommand.php
+++ b/src/pocketmine/command/defaults/TeleportCommand.php
@@ -93,6 +93,10 @@ class TeleportCommand extends VanillaCommand{
 
 			return true;
 		}elseif($target->getLevel() !== null){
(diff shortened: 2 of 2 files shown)


ZedCee:
Still seems to occur.  I am wondering if this is partially client side.

ZedCee:
It's actually a little worse off because now bed do not elapse time.

dktapps:
Welp

dktapps:
Different approach - disallow teleporting sleeping players. Please review again.

ZedCee:
Tested, seems to work +1

Shouldn't the failed teleportation display an error?  That will likely be a demanded addition.

dktapps:
OK, try that.

ZedCee:
Tested. Works. +1

svilex:
@dktapps this line should be moved or the player won't be teleported since is sleeping 

https://github.com/iTXTech/Genisys/blob/master/src/pocketmine/Player.php#L1235

dktapps:
@FaigerSYS TranslationContainer is useless in this case. That's only any use when the client has a matching translation string.

dktapps:
Okay, seems there's a lot more than just this bug that needs fixing for beds. I'll fix those issues, then reopen.

PR #1754, opened 2016-08-23
Added possibility to connect one genisys instance to many synapse/nemisys proxies

Tarik02:
### Description

This PR allows to connect one genisys to multiple synapse or nemisys instances.

### Reason to modify

I'm want to use more than one proxy for my server network, because one proxy cannot handle very many players.

### Tests & Reviews

I have tested the code and it works. This not breaks old configs.

### DIFF
--- a/src/pocketmine/resources/genisys_chs.yml
+++ b/src/pocketmine/resources/genisys_chs.yml
@@ -2,7 +2,7 @@
 
 #配置文件版本
 config:
- version: 18
+ version: 19
  
 level:
  #设置是否变换天气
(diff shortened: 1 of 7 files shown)


ishitatsuyuki:
1. Your formatting is messed up
2. Squash your commits
3. Multiple entrance for user doesn't make sense
-1

src/pocketmine/resources/genisys_chs.yml
```diff
@@ -97,18 +97,32 @@ redstone:
  #设置脉冲频率, 默认: 1s
  pulse-frequency: 1
 
+#TODO: Translate
 synapse:
- #如果你想使用 Synapse API 或连接到一个synapse服务器,请将enabled置为true
+ #If you want to use Synapse API or connect to a synapse server, enable this
```
dktapps: What's the point of doing this?

robske110: i think Chinese->English...

dktapps: Yes, in a CHINESE configuration file.

robske110: Lol

Tarik02: I'm not know chinese, but i'm want to update this config.

ishitatsuyuki:
I don't know what you're trying to do with this, but it doesn't make sense. You're also modifying things that shouldn't be changed. Closing.

PR #1770, opened 2016-08-24
Add client crash info

ABAxStorm:
### Description

This PR adds information to the README useful for those experiencing issues with immediately crashing upon join their server due to an unsupported block being placed.

### Reason to modify

Using BlockLauncher displays the blocks as an update image of update
### Tests & Reviews

I have used this method when I ran into the same issue before.


dktapps:
This sort of information should be in the wiki, not the README.

ABAxStorm:
I understand.

ZedCee:
This is a rather limited approach.  Only encompassing Android users.

ZedCee:
[MCEdit](https://github.com/Khroki/MCEdit-Unified) or [PMImporter](https://github.com/Muirfield/pmimporter) can also be used.

PR #1785, opened 2016-08-26
Fix floating text particle from steadfight2

RoofyManStar:
### Description

Floating Text Particle
### Reason to modify

FloatingTextParticle fix from SteadFight2
### Tests & Reviews

No,is not tested.. Probably work fine :)

https://github.com/Hydreon/Steadfast2/commit/c246afc27bc1fa095703fef0fe6b38f0beb4842b


dktapps:
Pointless modification and an obvious copy-paste job.

PR #1802, opened 2016-08-28
Add XBOX check function (new API)

FaigerSYS:
### Reason to modify

 
I think there must be some API for XBOX :)

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -221,6 +221,8 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 	protected $randomClientId;
 
+	protected $identityPublicKey;
+
 	protected $protocol;
 
 	protected $lastMovement = 0;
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
Please don't save the useless pubkey. It's only for encryption handshake.

Also, please make a specific reason to get if a user is authenticated. Since the PC edition doesn't has such a function, I didn't add it too.

FaigerSYS:
I want to make useful for some plugins. I have some ideas how to use this ability

ishitatsuyuki:
Then describe. Your idea is too abstract.

FaigerSYS:
It can be useful*

FaigerSYS:
Fot example I want to use it in my auth plugin

ishitatsuyuki:
There is already online-mode. More examples?

dktapps:
This is redundant,  since either all players _are_ authenticated, or all players are not.

dktapps:
It's pointless making per-player auth checks because it either applies to all players or none of them. A better approach is just to get the online-mode config from the server properties.

PR #1805, opened 2016-08-29
Synapse DataPacketSendEvent

Tarik02:
### Description

This fixes DataPacketSendEvent when using the synapse.

### Reason to modify

I'm want to use this event both with synapse.

### Tests & Reviews


dktapps:
Merge commits: nu-uh. And those methods do not have return values. -1.

PR #1809, opened 2016-08-29
0.16 packets IDs

svilex:
I don't think a description is needed for this


dktapps:
Mojang, you failures. Shift the whole list for one packet. 😂 Thanks very much.

SleepSpace9:
Now I understand why my Nemisys freezes since today after a short while. I guess it happens when a player with 0.16 joins Nemisys. No error output but the java process of Nemisys takes over 100% of CPU and nothing happens after that, no disconnect, but also no other processing. I don't have 0.16, so I can't test, but I'm convinced Nemisys will also need an update.

mhsjlw:
Protocol is version 90

daniktheboss:
Where do u get these packet from?

ishitatsuyuki:
@daniktheboss read the description, please don't spam on technical discussion.

PR #1812, opened 2016-08-30
New random number generator

ishitatsuyuki:
### Description

This is a solution to PM's sucked random algorithm. We're switching to xorshift, which is widely tested and used in many softwares.
### Reason to modify

This will solve the ultra-tall sugarcanes and cactus, especially on Android. It also resolves the peculiar issue on Android with the repeating terrain, grid-like positioning of objects, and various other issues.
### Tests & Reviews

Concerns: Maps that generated before will be broken due to the changed output of algorithm.

### DIFF
--- a/src/pocketmine/utils/Random.php
+++ b/src/pocketmine/utils/Random.php
@@ -21,11 +21,35 @@
 
 namespace pocketmine\utils;
 
-
 /**
- * Unsecure Random Number Noise, used for fast seeded values
+ * XorShift128Engine Random Number Noise, used for fast seeded values
+ * Most of the code in this class was adapted from the XorShift128Engine in the php-random library.
  */
 class Random{
+	const X = 123456789;
+	const Y = 362436069;
+	const Z = 521288629;
+	const W = 88675123;
+
+	/**
+	 * @var int
+	 */
+	private $x;
+
+	/**
+	 * @var int
+	 */
+	private $y;
+
+	/**
+	 * @var int
+	 */
+	private $z;
+
+	/**
+	 * @var int
+	 */
+	private $w;
 
 	protected $seed;
 
@@ -33,7 +57,7 @@ class Random{
 	 * @param int $seed Integer to be used as seed.
 	 */
 	public function __construct($seed = -1){
-		if($seed == -1){
+		if($seed === -1){
 			$seed = time();
 		}
 
@@ -44,9 +68,12 @@ class Random{
 	 * @param int $seed Integer to be used as seed.
 	 */
 	public function setSeed($seed){
-		$this->seed = crc32(pack("N", $seed));
+		$this->seed = $seed;
+		$this->x = self::X ^ $seed;
+		$this->y = self::Y ^ ($seed << 17) | (($seed >> 15) & 0x7fffffff) & 0xffffffff;
+		$this->z = self::Z ^ ($seed << 31) | (($seed >>  1) & 0x7fffffff) & 0xffffffff;
+		$this->w = self::W ^ ($seed << 18) | (($seed >> 14) & 0x7fffffff) & 0xffffffff;
 	}
-
 	public function getSeed(){
 		return $this->seed;
 	}
@@ -66,13 +93,15 @@ class Random{
 	 * @return int
 	 */
 	public function nextSignedInt(){
-		$t = ((($this->seed * 65535) + 31337) >> 8) + 1337;
-		if(PHP_INT_SIZE === 8){
-			$t = $t << 32 >> 32;
-		}
-		$this->seed ^= $t;
+		$t = ($this->x ^ ($this->x << 11)) & 0xffffffff;
 
-		return $t;
+		$this->x = $this->y;
+		$this->y = $this->z;
+		$this->z = $this->w;
+		$this->w = ($this->w ^ (($this->w >> 19) & 0x7fffffff)
+							 ^ ($t ^ (($t >> 8) & 0x7fffffff))) & 0xffffffff;
+
+		return $this->w;
 	}
 
 	/**


ZedCee:
Can't wait to test this out!  I'll be looking into it soon.  :cactus: :bamboo:

dktapps:
This isn't ready for merging yet. Setting seeds does not currently work.

ZedCee:
I personally did not notice any particular issues using /manyworlds and creating random seeds.

I did however find:
* Some odd lines and symmetry appears in certain large bodies of water, they seem to be presistant, but don't appear in all bodies.  I have logged on and off to find the same lines (intially thinking it was how the client handled the generating world).
image
image
image

* I was only able to find exactly one oversized Sugarcane, Cactus all appeared 3 or shorter. (Maybe two generated on top of each other, as I have seen it exactly once in vanilla play)
image

* There are a fair number of very tall birch, this however may just be part of the gen.
image

These were the general oddities I found.  They may not be directly related, but are more noticiable with this branch it seems.  Other than the above, it seems to function appropriately. The biomes are large, and seemingly diverse.  I can look into this further later today.

thebigsmileXD:
@ZedCee thats actually a client issue. should be fixed in 0.16 though, they changed water alot.
Tall birch trees are also kind of common in MCPE. Its a feature.
The issue with cacti and sugarcane is known. But your whole comment actually doesnt belong here ;) It has nothing to do with random numbers

ZedCee:
@thebigsmileXD My note is not whether or not the water is a client bug, but that it is FAR MORE prevalent on this branch, than from `master`.  With the odd coastal lines extending across full bodies of water more consistently as opposed to the single chunks found in `master` gen.  I felt this was due to a client issue, except that these lines extend over the full bodies of water.  
I also note, that I HAVE seen the tall sugarcane before, and confirmed this to be vanilla behaviour, in line with the wiki, with @ishitatsuyuki in chat.   
My comment regarding the birch is that it seems apparent that there is a rather high percentage more of them than prior.  This may be the expected behaviour and not in direct relation.

You may also note, if you read my post, I did not "minus one" the PR (the rejection was added by someone else)  and I would be testing later after I got sleep (quickly testing in the wee hours of the morning from my locale).   That from first look, these were the oddities I had noticed were more than usual.  
 I also did not experience seed issues with /manyworlds.  And from forward appearance, biomes were large and diverse; lacked previously notable symmetry in the land masses that may have been found before.  I was actually quite ready to "plus one" after confirming above suspicions of normal gen related anomalies.  As these were the only "very tall" things that I had found.

I have been working with iTXTech for the last couple months, generating at least one, if not, several maps per week.  For this round of testing I generated at least 6 separate maps/seeds (3 `master`, 3 `rng`).  These are the notable differences found on this branch.  I would presume that with your development on CS you would not actually have the time to make such tests on a Genisys server.  I apologize that I may have lacked clarity on some of these points, however I was falling asleep by the time I posted this and would be back to test more after further discussion with the rest the staff.

dktapps:
@ZedCee the PullApprove regex fails hard. Usually there'll be a negative 1 in the URL.

ZedCee:
Out of curiousity;  if applied to the `mobs` branch, would this not improve the selection of various slime sizes?

dktapps:
@ZedCee no. That's a bug long since fixed and doesn't use this generator.

ABAxStorm:
This will most likely make many server owners angry when they update and all of their worlds are broken.

ishitatsuyuki:
@ABAxStorm we don't know how many server really use this generator. Many server I saw from the issue tracker has bunch of plugins with a handbuilt map, or superflat thingy.

dktapps:
Pkease retest and feedback.

dktapps:
In testing I now observe that most of cacti and sugarcane are now 2 blocks tall, with an occasional 3 and a very rare 4. Terrain generation also now seems to yield caves more frequently.

src/pocketmine/utils/Random.php
```diff
@@ -21,19 +21,43 @@
...
 
 	protected $seed;
 
 	/**
 	 * @param int $seed Integer to be used as seed.
 	 */
 	public function __construct($seed = -1){
-		if($seed == -1){
+		if($seed === -1){
 			$seed = time();
```
ishitatsuyuki: Hmm.. Minecraft use "really secure bytes" iirc

dktapps: @ishitatsuyuki how do you mean? time() sucks admittedly

ishitatsuyuki: Uh, seems it's just the time, nvm

ishitatsuyuki:
+1

ZedCee:
I will retest when this PHP is compiled. :D

dktapps:
@ishitatsuyuki _You_ opened this PR, remember? You can't plus-one it :D

On your behalf I'll hit it though. Approved.

ZedCee:
I think your changes then superseded his.
I'm in the midst of testing.  I must say I'm seeing some nice things.

dktapps:
@ZedCee I made the commits, he opened the PR. Technically he's the author and cannot give an approve. :laughing:

ZedCee:
A minor note, aren't text based seeds supposed to be converted?
Or has this not been implemented?

dktapps:
@ZedCee I think they are, yes. This however is not relevant to the RNG.

ZedCee:
Well I can say without a doubt this brings improvements to the generation in general.
Cacti and sugarcane heights are great. (Still too many desert and beach cacti :| )

The biomes range much more beautifully between tiny residuals that appear in great places to utterly massive.  From the 3 tree spruce between the tundra and oak forest, seen again with slivers of birch here and there and vice-versa, or the awesome lone mountains and oak in the middle of a massive taiga woods.

There are "seas" now, bodies of water large enough to consider a boat.  Almost.
Caves seem present here and there as you mentioned as well.

And if the water fixes are coming in 0.16.0, I would have to note that using 0.15.7 I am finding zero water lines in the water.  No weird shadows on the ground.

#### OS and versions
* PHP: 7.0.10 (Arch)
* Server OS: Arch Linux
* Game version: 0.15.7 (iOS)

dktapps:
The string seeds issue is duly noted and will be investigated. Merging.

dktapps:
Does anyone happen to know where the original code for this came from? GitHub code search is yielding nothing useful.

PR #1814, opened 2016-08-30
Made Custom Enchantments possible

KairusDarkSeeker:
### Description
I've made this pull request to make CustomEnchantments possible for a plugin

### Reason to modify

Yes
Yes
I changed private __construct to a public function and changed protected static $enchantments to a public static variable

### Tests & Reviews
I have tested the code and it works perfectly with a custom enchants plugin

### DIFF
--- a/src/pocketmine/item/enchantment/Enchantment.php
+++ b/src/pocketmine/item/enchantment/Enchantment.php
@@ -134,7 +134,7 @@ class Enchantment{
 
 
 	/** @var Enchantment[] */
-	protected static $enchantments;
+	public static $enchantments;
 
 	public static function init(){
 		self::$enchantments = new \SplFixedArray(256);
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
I think this is barely needed and considered as anti-pattern. -1

KairusDarkSeeker:
@ishitatsuyuki why?

ishitatsuyuki:
Explain what you are **really** trying to do instead of concepts. The reason to modify is insufficient.

KairusDarkSeeker:
@ishitatsuyuki I made the PR to make the plugins able to access the Enchantments then all people will now be able to create a custom enchantment plugin by not making a Modder genisys server phar file :)

ishitatsuyuki:
Well, a foolish idea. They won't work without server modification, not only this class.

KairusDarkSeeker:
@ishitatsuyuki okay...............................................................

PR #1848, opened 2016-09-04
Feature update: Add ability to enable/disable commands in pocketmine.yml

dktapps:
### Description
This PR adds the ability to enable or disable commands in pocketmine.yml, allowing server owners to declutter the /help command of useless commands.

### How this works
It's quite straightforward. Under the new `commands` section in pocketmine.yml there is a default behaviour option. This will be used when no specific configuration is set for a command. Setting the default as FALSE will provide a whitelist behaviour (only the commands you select are enabled). Setting the default as TRUE will provide a blacklist behaviour (remove commands you don't want).

To set specific behaviour for a command, simply add a key under the `commands` section with the name of the command you want to override the default for. For example, using the whitelist behaviour and wanting to disable the /version and /plugins  commands, I would do this:
```
commands:
 #Here you can customize server commands
 #Specify command names to override the default set here.
 #If no custom value is defined for a command, the default will be used.
 #NOTE: Some commands cannot be disabled here, such as the important ones like /stop, /reload, etc.
 default: true
 #Set override values per command here
 #For example, uncommenting the below will disable /plugins and /version
 version: false
 plugins: false
```

### Reason to modify

Add ability for server owners to remove commands that are never used. For example, I never use /bancid, but no doubt it comes in useful for other people. 
This works by preventing disabled commands from being loaded.

#### Things worth noting:
- This does NOT block commands; it prevents disabled commands from being loaded at all. This is therefore less flexible than using a permissions plugin as it cannot be changed at runtime. However, if all you want to do is remove some commands that you never use, then this will be a bonus.
- Some commands cannot be disabled for obvious reasons, such as /help, /stop, /reload

### Tests & Reviews
I have tested this under both white- and blacklist configuration and found no issues. Please comment below if you encounter issues.

### Please review things below:
- This configuration will also affect plugin commands. Whether or not this behaviour should be retained is something I'm unsure on. Please comment.

### DIFF
--- a/src/pocketmine/command/SimpleCommandMap.php
+++ b/src/pocketmine/command/SimpleCommandMap.php
@@ -86,12 +86,19 @@ class SimpleCommandMap implements CommandMap{
 	 * @var Command[]
 	 */
 	protected $knownCommands = [];
+	
+	/**
+	 * @var bool[]
+	 */
+	protected $commandConfig = [];
 
 	/** @var Server */
 	private $server;
 
 	public function __construct(Server $server){
 		$this->server = $server;
+		/** @var bool[] */
+		$this->commandConfig = $this->server->getProperty("commands");
 		$this->setDefaultCommands();
 	}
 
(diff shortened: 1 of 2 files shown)


Muqsit:
gg +1
Bukkit servers generate a help file (help.yml) for modification of  the '/help' command. Hope we can get something of that kind in the near future.

XFuryMCPE:
Awesome! Also, should probably add an option to change the Unknown command message :D

dktapps:
@muqsit Disabled commands will not appear in /help, that's the main reason I added this.

@XFuryMCPE any suggestions for an alternative message?

PeratX:
+1

XFuryMCPE:
If a server has a specific tag they use in their server, or a modified /help command, they could do "Error> Command not found! Type /commands for a list of available commands!" or something like that

daniktheboss:
Alternative... have an option for people to write in it 😁

PR #1859, opened 2016-09-05
Fix #1857

### LINKED ISSUE #1857: Paintings can be killed even the damage event is cancelled
I coded a plugin which should cancel every damage event.
```
public function onEvent(EntityDamageEvent $event){
$event->setCancelled();
}
```
Still, paintings will get removed.
May be cause of the attack() override.
photo_2016-09-05_01-30-47

thebigsmileXD:


dktapps:
Looks good. +1

SOF3:
And are you gonna do this for every single event trigger that involve subsequent use of player functions?

dktapps:
Actually, your changes are somewhat redundant. Returning when the event is cancelled will do nothing whatsoever as there is no code after that condition. Removing the close() should be sufficient. -1

PR #1864, opened 2016-09-05
Fix #1852

### LINKED ISSUE #1852: Only first argument of stop reason displayed
if the command stop is issued with extra args, it will only stop the server like /stop be right back -> "be"

Solution: implode(" ", $args) in the stop command.

thebigsmileXD:


thebigsmileXD:
New usage: `/stop force this is the reason` -> stops server and restarts it, shows reason
`/stop this is the reason` -> stops the server and shows the reason
`/stop` -> stops the server with default message
`/stop force` -> stops the server and restarts it, displays the default reason.

dktapps:
I committed this as your credentials directly to ditch those ugly merge commits. Thanks. https://github.com/iTXTech/Genisys/commit/afcbcb8b1156fe3cdd3f5e4d739df99ad99b50c9

PR #1867, opened 2016-09-05
Fixed Player Closed in Events

### LINKED ISSUE #1828: "Fake" player count
#### Issue description

I always get the players bug whenever i leave my server up for 1 day or more. It says server is full although its only 30/70 people, the player list is full of offline players.

Don't actually know. Just leave your server for some days. (I don't think this is a nice reproduce way lol)

#### OS and versions

* Genisys: 1.2dev-24_c19e850
* PHP: 7.3 (?)
* Server OS: CentOS 7
* Game version: 0.15.x

```
```

### LINKED ISSUE #1865: Bad fix of hasPermission() on null
#### Issue description

Currently, when a plugin kicks a player in PlayerLoginEvent, etc., the server will continue to execute code that handles the client without checking whether the player is online. While there are no errors, it is _WRONG_ for the server to continue doing stuff on a player that isn't online, assuming it is online.

Previously, this problem was mainly reflected by the internal call on `Player::hasPermission` that leads to a crash. Although an explicit null check is done on hasPermission, calling `hasPermission` on a closed player should indeed raise a warning (through throwing an exception) rather than returning a boolean false, as some plugins may hence cache and assume that the answer is _`false`_, while the actual value is _undefined_.

Vice versa. This also applies to myriad event calls related to players. Refer to comments below.

#### Possible resolution
1. Override `Event::isCancelled` in `PlayerEvent` (and probably `EntityEvent` too), and add a check to confirm that the player is online:

```php
public function PlayerEvent::isCancelled() : bool{
  return parent::isCancelled() and $this->player->isOnline();
}
public function EntityEvent::isCancelled() : bool{
  return parent::isCancelled() and !($this->entity instanceof Player and !$this->entity->isOnline());
}
```

2. Override the new override methods like `PlayerEvent;:isCancelled` for events that should not implement such check, such as `PlayerQuitEvent`:

```php
public function PlayerQuitEvent::isCancelled() : bool{
  return Event::isCancelled();
}
```

#### Side effects
Slight backwards incompatibility. We cannot ask older plugins that declare events like `PlayerQuitEvent` to override `isCancelled`.

#### Resolution of side effects
I have been unable to come up with any resolution of side effects.

However, according to @shoghicp, the main author of the PocketMine event and plugin system, plugins [_SHOULD NOT_](https://tools.ietf.org/html/rfc2119) extend PlayerEvent, etc. All events declared by plugins should extend `PluginEvent`. (It is believed that an exception is plugin events that share same handler lists as default events are not included by this statement, although this is not a good idea either)

For old plugins that extend `PluginEvent`, they are totally not covered by the proposed change in the _Possible Resolution_ section above. The issues with them may or may not continue to exist, but they will only have the chance to get better, not worse, anyway.

Write a plugin that kicks a player in `Player(Pre)?LoginEvent` without cancelling the event.

#### OS and versions

* Genisys: e85b6f14436b3b238f359d8ffcea4af8d1118892
* PHP: Doesnt' matter
* Server OS: Doesn't matter
* Game version: Doesn't matter

Plugin used to produce the issue.
```php
<?php

/**
 * @name KickOnJoin
 * @author SOFe
 * @version 1.0.0
 * @api 1.0.0
 * @main KickOnJoin
 */

class KickOnJoin extends \pocketmine\plugin\PluginBase implements \pocketmine\event\Listener{
	public function onEnable(){
		$this->getServer()->getPluginManager()->registerEvents($this, $this);
	}
	public function onLogin(\pocketmine\event\player\PlayerPreLoginEvent $event){
		$event->getplayer()->close("Test");
	}
}
```

dschwartz783:
### Description
Fixes #1828 and partially fixes #1865 

### Reason to modify
Because #1828  needed to be fixed...

### Tests & Reviews
Definitely fixes #1828

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2214,6 +2214,10 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 			return;
 		}
+		
+		if(!$this->isConnected()){
+			return;
+		}
 
 		$pk = new PlayStatusPacket();
 		$pk->status = PlayStatusPacket::LOGIN_SUCCESS;
@@ -2440,8 +2444,10 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 					break;
 				}
-
-				$this->onPlayerPreLogin();
+				
+				if($this->isConnected()){
+					$this->onPlayerPreLogin();
+				}
 
 				break;
 			case ProtocolInfo::MOVE_PLAYER_PACKET:


SOF3:
This only addresses one of the almost infinite problems referred to in #1865.

dktapps:
I believe this mainly targets #1828 actually.

SOF3:
@dktapps According to my tests with a slightly-modified [Specter](https://githbu.com/Falkirks/Specter) plugin and the script plugin provided in #1865, the output is like this:

```
$ s s 1dum
2016-09-05 [18:22:01] [Server thread/DEBUG]: Attempted to drop a null crafting inventory
2016-09-05 [18:22:01] [Server thread/INFO]: 1dum[/SPECTER:19133] logged out due to generic reason
2016-09-05 [18:22:01] [Server thread/INFO]: 1dum[/SPECTER:19133] [ClientID: 1] logged in with entity id 1 at (world, 0, 0, 0)
2016-09-05 [18:22:01] [Server thread/INFO]: Session started.
$ list
2016-09-05 [18:22:07] [Server thread/INFO]: There are 0/20 players online:
2016-09-05 [18:22:07] [Server thread/INFO]:
```
(obviously I added the `$` signs)

Therefore I believe that #1828 isn't related to this problem.

dschwartz783:
Actually, #1865 only mentions PlayerLoginEvent in the description. This is for PlayerPreLoginEvent. I suppose it isn't really related...

SOF3:
@dschwartz783 sorry, my wrong. I edited the issue.

dschwartz783:
Hmm, you're right @dktapps. This does indeed fix that issue too.

dktapps:
@sof3 If you read #1828, his server says 30 players online when it's actually full. That's what I was referring to. I believe this is the bug that causes it.

dktapps:
This looks fine. +1

Muqsit:
+1

PR #1869, opened 2016-09-05
Fix https://github.com/iTXTech/Genisys/issues/1857

thebigsmileXD:
### Description

Same like before, 100% tested. Only if the even't is not cancelled, particles play and the painting item drops.

### Reason to modify

Old PR was missing a "false"
### Tests & Reviews

Tested with simple entity damage cancelling plugin.

### DIFF
--- a/src/pocketmine/entity/Painting.php
+++ b/src/pocketmine/entity/Painting.php
@@ -28,8 +28,9 @@ class Painting extends Hanging{
 
 	public function attack($damage, EntityDamageEvent $source){
 		parent::attack($damage, $source);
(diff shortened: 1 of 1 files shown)


dktapps:
I don't think paintings have destroy particles actually.  But other than that looks okay.

thebigsmileXD:
i've sendt a vido into the chat group. The particles are correct, the sounds aren't 100% perfect, but close enough

PeratX:
+1

dktapps:
+1 and merge.

PR #1874, opened 2016-09-05
Generator: Fix Void generator

dktapps:
### Description
Fix the SkyGrid fail in the Void generator.

### Reason to modify
- Generate a "void" world with the current generator. Result = a grid of grass blocks at about y = 64 and players plummet to their deaths on spawn.
- With my generator, only ONE grass block is generated - in the spawn chunk. This provides a place to stand and players will spawn on this block. It also provides something to build against.

This works by checking if the spawn coordinates are within the chunk we are currently generating. If yes, we add a grass block. If no, we cache a copy of the blank chunk we created and store it so that we don't need to continually regenerate empty chunks. This should improve generation efficiency.

### Tests & Reviews

Before:
screenshot_20160905-235601

After:
screenshot_20160905-235339

### DIFF
--- a/src/pocketmine/level/generator/Void.php
+++ b/src/pocketmine/level/generator/Void.php
@@ -36,6 +36,8 @@ class Void extends Generator{
 	/** @var Random */
 	private $random;
 	private $options;
+	/** @var FullChunk */
+	private $emptyChunk = null;
 
 	public function getSettings(){
 		return [];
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
Good, lol, too much work for just a void generator, I don't know if caching makes sense in PHP but I think it's unneeded.

Anyway, +1

ZedCee:
It's an improvement from the odd grass grid.

Don't mind me if I stick with [EmptyWorld](https://github.com/Muirfield/EmptyWorld) for the automatic biome settings and platform customization reminiscent of MCPC.

Tested, works better than the existing +1

PR #1877, opened 2016-09-06
Added Missing Comment

dschwartz783:
### Description
just added a missing comment I noticed. Nothing major... Should be there.

### Reason to modify
Comments matter to IDEs, and should matter to you too! 🍡 (I was going to do :D, but then I saw dango....)

### Tests & Reviews

It's a comment. You don't test comments...


ZedCee:
Probably good to comment obscure code +1

dktapps:
I was going to merge directly, but since we have a +1 already I'll follow suit. Approved.

PR #1882, opened 2016-09-06
DataPacketSendEvent with Synapse now works

Tarik02:
### Description

This fixes DataPacketSendEvent with synapse.

### Reason to modify

I'm use synapse and i'm want to use DataPacketSendEvent, but this event not works with synapse.

### Tests & Reviews


PeratX:
This will slow down the synapse.

Tarik02:
@PeratX, this event used in pure genisys and this not slow down it a LOT.

Tarik02:
@PeratX, please, reopen this PR.

Issue #1709, opened 2016-08-20
disable-log issue


FaanMaario:
#### Issue description

I don't know if anybody have the same issue, but for me with the "disable-log=_true_", my server doesn't boot at all and give me an error.
And when I set this to "_false_", my server works fine.

#### Steps to reproduce the issue

1) Open genisys.yml
2) set the "disable-log" to "_true_"
3) reboot your server
4) an error will appear and your server doesn't boot

#### OS and versions

* Genisys: 19c0136e
* PHP: PHP7 w/ OpenSSL
* Server OS: Debian 8
* Game version: 0.15.6

#### Crashdump, backtrace, memory dumps, plugins list or other files

```
zend_mm_heap corrupted
```

ishitatsuyuki:
Have you tried multiple times? This can occur regardless if log is enabled or not.

FaanMaario:
Already tested on multiple servers, and it's the same issue.

TDMidkwhatisxd:
I use latest , and no errors !
i always disamble  log btw!

FaanMaario:
@TDMidkwhatisxd I have now the latest, and the same issue is here.

gaoyichuan:
Zend crash is a common issue, and it's due to php itself, nothing to do with Genisys

FaanMaario:
@gaoyichuan But how to fix it and why it appears when I disable the logs?

ishitatsuyuki:
This is definitely caused by PM's poor syncing between threads and the corruption as a result. We shouldn't blame everything on pthreads.

mote166:
just disable it . almost every server disables it (mine too)

dktapps:
Closing due to not reproducible and lack of activity. Incidentally: `true`, not `on`.

ishitatsuyuki:
Reopening. Uh, @dktapps - this is 100% reproducible with Docker container. I'm running Linux and i7-6500U.

dktapps:
Ah, my bad.

ZedCee:
I can reproduce with here too
#### OS and versions
* Genisys: https://github.com/iTXTech/Genisys/commit/79d7093d2c5ad643a8402fae34f80d7e0600d941  (git pull /src)
* PHP: 7.0.9 (Arch)
* Server OS: Arch Linux
* Game version: 0.15.6

FaanMaario:
Yes, I made a small typo (I write "true" not "on/off") sorry ^^

[...]

PR #1883, opened 2016-09-06
Safer method for disabling log at runtime

dktapps:
### Description
This **may** fix the issue outlined in #1709. 

### Reason to modify
Previously, the logger writing to file is enabled by default, and will log a couple of lines before it is disabled. Now, I'm not 100% sure what's going on, but the `synchronized` combined with the fact that the logging to file will be unceremoniously broken off when the log is disabled by config led me to this.

This is a cleaner solution which only affects the actual file content writing. File write is also off by default, so it will only start when the config is loaded.

### Tests & Reviews

I have tested the code and it doesn't crash or cause any noticeable issues. However,  I cannot verify whether it fixes the aforementioned issue or not. Please test and review.

### DIFF
--- a/src/pocketmine/utils/MainLogger.php
+++ b/src/pocketmine/utils/MainLogger.php
@@ -37,7 +37,7 @@ class MainLogger extends \AttachableThreadedLogger{
 	private $consoleCallback;
 
 	/** Extra Settings */
-	protected $write = true;
+	protected $write = false;
 
 	public $shouldSendMsg = "";
 	public $shouldRecordMsg = false;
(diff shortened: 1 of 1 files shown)


XFuryMCPE:
Gg

XFuryMCPE:
Now add the same setting to Nemisys

dktapps:
@XFuryMCPE that isn't a solution to the real problem, which is that Nemisys is leaking file references.

ZedCee:
Does not appear to crash the server on start anymore.  Logs can mostly be disabled, albeit it still produces the empty `server.log`.  Of no consequence for size, but some may not like the clutter.

Tested +1

PeratX:
+1

PR #1891, opened 2016-09-07
Updated tur.ini

SuperAdam47:


dktapps:
No description provided and I don't see any reason to delete all that. -1

PR #1917, opened 2016-09-10
Add API for XBOX.

robske110:
### Description
This PR adds some abilities for plugins. They can now check if a player is Authenticated with XBOX and also get the identity key for recognising that exact player. It also opens up the chainData array.

### Reason to modify

Plugins lacked the ability to get any XBOX information. And also we open plugins the full chainData now, so they can get to data they want which is not (yet) used in Genisys.

### Tests & Reviews

Please review things below:
- [ ] Do the function names fit?
- [ ] Does the code follow a good style?

Cheers!

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -233,6 +233,8 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	/** @var Vector3 */
 	protected $sleeping = null;
 	protected $clientID = null;
+	protected $authenticated = null;
+	protected $chainData = [];
 
 	private $loaderId = null;
 
@@ -445,8 +447,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 	/**
 	 * This might disappear in the future.
-	 * Please use getUniqueId() instead (IP + clientId + name combo, in the future it'll change to real UUID for online
-	 * auth)
+	 * Please use getUniqueId() (IP + clientId + name combo) or getIdentityPublicKey() (Xbox key) instead!
 	 */
 	public function getClientId(){
 		return $this->randomClientId;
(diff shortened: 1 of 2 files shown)


ishitatsuyuki:
Public key is a junk and is only used in the first step of key exchange. **Don't store it.**

Also, there has been a controversy about this. See #1802.

robske110:
Ok. Which key is not junk?
In #1802 dktapps says that either all or no players are not authenticated, which is just wrong. 
I'll remove the identity key storing, and dig into it to find a key which is not total junk.

ishitatsuyuki:
Only the key used in actual encryption has meaning.

Think: does the boolean has any other meanings? Server owners use their hammers regardless if they're authed, I don't think they really make sense.

robske110:
getChainData() opens a whole lot of information flow to plugin developers.
rebase, updated commit description

ishitatsuyuki:
The chain data does nothing, actually. If you just want the XUID, it's not intended to used by plugins, since the UUID is already unique for Xbox.

robske110:
wait, the chain data does contain everything... (not the protocol though)
It's just a raw way to just get some data, we even might not use...

dktapps:
The reason I said that was redundant was because if Xbox auth is not enabled, the server will not authenticate _any_ players. So you can tell if players are being authenticated by checking if online-mode is enabled. Either auth is turned on and they are _all_ authenticated to XBL, or authentication is turned off, in which case _no_ players are authenticated anyway.

robske110:
The function returns if the player is using a valid xbox login or not

robske110:
Independent from online mode enabled or not

dktapps:
@robske110 that's my point. If online-mode isn't enabled it isn't verified anyway.

robske110:
It is.

MCPEPIG:
This is useful for plugin devs.

MCPEPIG:
For example, if you want to stop non-xbox users from chatting you can use this api.

robske110:
They will never accept it because they just don't care about plugins.

ishitatsuyuki:
It simply doesn't make sense. Unauthenticated players can fake anything, thus they can get no trust. And is muting them a good idea? I don't feel so.

MCPEPIG:
It was just an example. And blocking most non xbox users is betetr than blocking none 😉

PR #1920, opened 2016-09-10
Fix a bug with updateAround only updating one block. Close #1618

### LINKED ISSUE #1618: updateAround not work properly
#### Issue description

updateAround not work properly

1. put a Jungle wood
2. put 4 cocoBean around this wood.
3. break the Jungle wood.
You will find that only one CocoBean update properly.
#### OS and versions

* Genisys: 3c07347286e7cbca1ec38a695ef903968fa83dd7
* PHP:7
* Server OS:Win10
* Game version: MCPE0.15.2

Reference:
 - 1 #1599
 - 2 https://github.com/PocketMine/PocketMine-MP/commit/781de3efabcd526281fd1adc8ce881d408471563
 - 3 https://github.com/PocketMine/PocketMine-MP/pull/4024

dktapps:
### Description
Fix #1618 - updateAround only updates one block.

### Reason to modify
- See #1618 for method to reproduce this bug.

As best I can understand, updateAround was being supplied with a floating-point Vector3, which when offset by +/-1, caused rounding errors resulting in the same block being updated several times. This fixes the issue by flooring the Vector3 before updating anything.

### Tests & Reviews
I have tested the code and it works flawlessly.
Tests:
 - Place a block and attach torches to all sides, then break the block. The torches will now all be updated.

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -1115,6 +1115,7 @@ class Level implements ChunkManager, Metadatable{
 	 * @param Vector3 $pos
 	 */
 	public function updateAround(Vector3 $pos){
+		$pos = $pos->floor();
 		$this->server->getPluginManager()->callEvent($ev = new BlockUpdateEvent($this->getBlock($this->temporalVector->setComponents($pos->x, $pos->y - 1, $pos->z))));
 		if(!$ev->isCancelled()){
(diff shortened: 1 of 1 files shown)


Muqsit:
Nice find lol +1

ishitatsuyuki:
Good lol, why didn't this exist.
+1

dktapps:
Blame @PEMapModder ¯\_(ツ)_/¯

PR #1942, opened 2016-09-15
Add warning for xdebug on startup

dktapps:
### Description
I intended to commit this directly, but either GitHub has screwed up the branch protection or somebody has tampered with the settings, because I can't commit to master.

This PR adds a warning for xdebug on startup.

### Reason to modify
- xdebug has a huge impact on runtime performance. Having it loaded and not realising could cause server owners lots of grief.

### Tests & Reviews


PeratX:
+1

PR #1944, opened 2016-09-15
Fixed start.cmd on windows, when using system PHP.

Tarik02:
### Description

I'm tried to clone Genisys and just run it. I'm get this error:
```
[CRITICAL] Unable to find the pthreads extension.
[CRITICAL] Please use the installer provided on the homepage.
```

I'm has pthreads in system PHP, but if use `-c bin\php` when this folder isn't exists, php will not use extensions.

### Reason to modify

Fix this error.

### Tests & Reviews


ishitatsuyuki:
uh, I don't think this really makes sense. Either add it to global config/registry, or hack locally there.

ishitatsuyuki:
Uh, I'm going to close this. Feel free to post your argument.

PR #1946, opened 2016-09-16
Player: Replaced food system with PM's food system

PeratX:
### Description

Food system is **awful**. So I replaced it with PM's food system.

### Reason to modify

* Removed some deprecated methods
* Improved Cake
* Added Mining, attack, damage cause for food

### Tests & Reviews

* Tested (stay under water to get hurt and back to ground, regain health and food level decreased)


Muqsit:
setFood and getFood gone?
It might be a good function for plugin developers. But great job 👍

PeratX:
@Muqsit You can find these methods in Human.php

dktapps:
XD - "Throw out our mess, use PM's instead" - a growing trend, I like it. Will test this later.

dktapps:
So, 1.9.2 lasted a long tine then 😆

ishitatsuyuki:
You're changing too much for doxygen. Remove it for now, I will push mine later to other branch.

PeratX:
You can change it on my base.

XFuryMCPE:
gg. Genisys is slowly becoming better and better

src/pocketmine/entity/Human.php
```diff
@@ -529,38 +529,40 @@ public function entityBaseTick($tickDiff = 1, $EnchantL = 0){
...
 
-		/*$food = $this->getFood();
-		$health = $this->getHealth();
-		if($food >= 18){
-			$this->foodTickTimer++;
-			if($this->foodTickTimer >= 80 and $health < $this->getMaxHealth()){
-				$this->heal(1, new EntityRegainHealthEvent($this, 1, EntityRegainHealthEvent::CAUSE_SATURATION));
-				$this->exhaust(3.0, PlayerExhaustEvent::CAUSE_HEALTH_REGEN);
-				$this->foodTickTimer = 0;
+		if($this->getLevel()->getServer()->foodEnabled){
```
dktapps: Why not use server directly instead of adding extra middlemen?

Muqsit:
GG @PeratX =D

src/pocketmine/entity/Human.php
PeratX: It's the same as Server::getInstance()

dktapps: It's unneeded, you can access $server directly.

PeratX:
Ready for squash and merge.

src/pocketmine/entity/Human.php
```diff
@@ -529,7 +529,7 @@ public function entityBaseTick($tickDiff = 1, $EnchantL = 0){
 		}
 		$hasUpdate = parent::entityBaseTick($tickDiff, $EnchantL);
 
-		if($this->getLevel()->getServer()->foodEnabled){
+		if($this->server->foodEnabled){
```
dktapps: +1

src/pocketmine/Player.php
```diff
@@ -3123,8 +3061,21 @@ public function handleDataPacket(DataPacket $packet){
 				$this->setDataFlag(self::DATA_FLAGS, self::DATA_FLAG_ACTION, false); //TODO: check if this should be true
 
 				switch($packet->event){
-					case 9: //Eating
-						$this->consumeHeldItem();
+					case EntityEventPacket::USE_ITEM: //Eating
+						$slot = $this->inventory->getItemInHand();
+
+						if($slot->canBeConsumed()){
```
SOF3: This check is unnecessary. It is unreasonable that this happens, and if this really happens, just let the plugin handle it (default cancelled, as on line 3070)

src/pocketmine/item/Item.php
```diff
@@ -821,6 +821,17 @@ public function setCount(int $count){
 		return $this->canBePlaced();
 	}
 
+	public function canBeConsumed() : bool{
```
SOF3: This function can be made `protected`, since only `canBeConsumedBy` should be called externally.

src/pocketmine/item/Item.php
```diff
@@ -821,6 +821,17 @@ public function setCount(int $count){
...
+		return false;
+	}
+
+	public function canBeConsumedBy(Entity $entity) : bool{
```
SOF3: Consider accepting a string value (class name) as well.

src/pocketmine/item/Potion.php
```diff
@@ -141,7 +141,7 @@ public function canBeConsumed() : bool{
 		return $this->meta > 0;
 	}
 	
-	public function canBeConsumedBy($entity) : bool{
+	public function canBeConsumedBy(Entity $entity) : bool{
 		return $entity instanceof Human;
```
SOF3: What about `Witch`?

src/pocketmine/item/Item.php
ishitatsuyuki: PM's access modifiers were never proper, let's not touch it.

SOF3: My point is not about the modifier. I'm talking about how it should be used. Even if they are improper, you can still add an `@internal` JavaDoc tag **if it is useful**.

src/pocketmine/item/Potion.php
dktapps: TODO ;)

PeratX: Please take a look at origin PocketMine-MP's code.

SOF3: @PeratX Please take a look at the author of original PocketMine-MP's code.

PR #1953, opened 2016-09-18
0.15.8 version update

andris155:
### Description

0.15.8 version update

### Reason to modify

Mojang updated Minecraft PE.


dktapps:
We update the version when a breaking change requires updates to the protocol. The protocol has not changed since 0.15.4.

robske110:
I would just set it to 0.15.x.

dktapps:
@robske110 you can't. The client doesn't like it when you do that.

robske110:
Only the _NETWORK constant. CS has that at 0.15. the other string accepts anything

dktapps:
We keep them the same for consistency,

robske110:
this is genisys

PR #1966, opened 2016-09-21
Update 0.16 support notice with link to issue

sekjun9878:
### Description
README Improvement

### Reason to modify
Better usability for readers

### Tests & Reviews
Testing is not necessary, use preview function.


dktapps:
👍 Thanks

PR #1991, opened 2016-09-27
Timezone related fix

iPocket:
### Description
Timezone related fix, with shoghi...
#BLAMESHOGHI

### Reason to modify
Fix an always false return value.

### Tests & Reviews
I have not tested the code.


KairusDarkSeeker:
👏 Untested PR and merged PR

dktapps:
@KairusDarkSeeker I tested it myself, get your facts straight.

PR #1999, opened 2016-09-29
Install script for mack linux .....

majovec:
### Tests & Reviews

 Tested on Debian 8 64bit

How to use :
1. download this scrip and upload on your folder.
2. open putty or console and write chmod 777 install.sh
3. wrote in console ./install.sh 
4. Have a fun

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+# Genisys Installation Script for Mac OS X and Linux(master)
+#  _    _            _    _______ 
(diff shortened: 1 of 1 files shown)


daniktheboss:
Kinda see no point. Raspberry pi just install Linux version.... anything else install bin, download phar. Done

dktapps:
We only support Docker installation for Linux, and I don't think mixing licences is a good idea. If you want to run a non-sandboxed installation it's recommended to compile your own PHP anyway from latest source.

PR #2000, opened 2016-09-29
add isFirstLogin function to synapse player

XFuryMCPE:
### Description

Adding the isFirstLogin function to synapse player

### Reason to modify

This will allow us to check whether it is the player's first login to the Synapse/Nemisys proxy or not. This would work well for global login plugins or some other things

### Tests & Reviews

Common sense says this works, but just letting you know I have not tested it.


XFuryMCPE:
Hey you said it was fine for tiny edits,

dktapps:
I'm trying to discourage you. I wouldn't have merged if there was an issue.

PR #2006, opened 2016-10-01
Make template more descriptive

legoboy0215:
### Description

I simply made the template more descriptive.

### Reason to modify

Nothing will break. I did this, well, because some things are not that descriptive on the template.

### Tests & Reviews
I reviewed it. English is not my native tongue, so correct me if I am wrong!


ishitatsuyuki:
Meaning less. Closing.

PR #2018, opened 2016-10-04
Fix dropped item disappearing when event cancelled

### LINKED ISSUE #1961: PlayerDropItemEvent not working
**Issue description**
https://youtu.be/RhHb93dPuvA
Pls watch video, item will be lose when i cancalled PlayerDropItemEvent
`public function noDrop(PlayerDropItemEvent $event) { $event->setCancelled(); } `

**OS and versions**

Genisys: 5e4bfc0
PHP Version: 7.0.3
OS : Linux, linux
Game Verison: 0.15.x

Sorry my bad english

XFuryMCPE:
### Description

Fix #1961

### Reason to modify

Read above

### Tests & Reviews

I don't have a testing device right now, please test for me.


dktapps:
This won't work. The item is not actually in your inventory anymore by the time this is called, it's in your FloatingInventory.

PR #2049, opened 2016-10-09
small edit

ondraczsk:
edit Pocketmine to Genisys


Muqsit:
Its called 'pocketmine.yml' but anyways...this is unneeded.

PR #2051, opened 2016-10-09
FloatingText as entity

### LINKED ISSUE #1986: Floating Text in all worlds
#### Issue description

If i made a plugin with floating text in 1 worlds than its in all worlds and all server with synapse but i havent this plugin on all servers...
#### OS and versions

- Genisys: 147
- PHP: 7.0.11
- Server OS: Ubuntu 14.,0
- Game version: 0.15.9

```
```

jjplaying:
### Description

It registers FloatingText as an entity.
### Reason to modify

It fixes #1986, you can spawn and despawn FloatingTexts like an entity.
### Tests & Reviews

I spawned and despanwed a FloatingText entity, also changed the world, and everything works.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -42,6 +42,7 @@ use pocketmine\entity\Enderman;
 use pocketmine\entity\Entity;
 use pocketmine\entity\FallingSand;
 use pocketmine\entity\FishingHook;
+use pocketmine\entity\FloatingText;
 use pocketmine\entity\Ghast;
 use pocketmine\entity\Human;
 use pocketmine\entity\Husk;
@@ -2936,6 +2937,7 @@ class Server{
 		Entity::registerEntity(Enderman::class);
 		Entity::registerEntity(FallingSand::class);
 		Entity::registerEntity(FishingHook::class);
+        Entity::registerEntity(FloatingText::class);
 		Entity::registerEntity(Ghast::class);
 		Entity::registerEntity(Husk::class);
 		Entity::registerEntity(IronGolem::class);
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -2936,8 +2936,8 @@ class Server{
 		Entity::registerEntity(Egg::class);
 		Entity::registerEntity(Enderman::class);
 		Entity::registerEntity(FallingSand::class);
-        Entity::registerEntity(FishingHook::class);
-        Entity::registerEntity(FloatingText::class);
+		Entity::registerEntity(FishingHook::class);
+		Entity::registerEntity(FloatingText::class);
 		Entity::registerEntity(Ghast::class);
 		Entity::registerEntity(Husk::class);
 		Entity::registerEntity(IronGolem::class);
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -2936,7 +2936,7 @@ class Server{
 		Entity::registerEntity(Egg::class);
 		Entity::registerEntity(Enderman::class);
 		Entity::registerEntity(FallingSand::class);
-		Entity::registerEntity(FishingHook::class);
+        Entity::registerEntity(FishingHook::class);
         Entity::registerEntity(FloatingText::class);
 		Entity::registerEntity(Ghast::class);
 		Entity::registerEntity(Husk::class);
--- a/src/pocketmine/entity/FloatingText.php
+++ b/src/pocketmine/entity/FloatingText.php
@@ -8,7 +8,7 @@ use pocketmine\Player;
 use pocketmine\utils\UUID;
 
 class FloatingText extends Entity {
-    const NETWORK_ID = 80;
+    const NETWORK_ID = 1000; // Just a random unused ID
 
     protected $title;
     protected $text;
--- a/src/pocketmine/entity/FloatingText.php
+++ b/src/pocketmine/entity/FloatingText.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace pocketmine\entity;
+
+use pocketmine\item\Item as ItemItem;
+use pocketmine\network\protocol\AddPlayerPacket;
+use pocketmine\Player;
+use pocketmine\utils\UUID;
+
+class FloatingText extends Entity {
+    const NETWORK_ID = 80;
+
+    protected $title;
+    protected $text;
+
+    public function getName(): string {
+        return "FloatingText";
+    }
+
+    public function setTitle($title) {
+        $this->title = $title;
+    }
+
+    public function setText($text) {
+        $this->text = $text;
+    }
+
+    public function spawnTo(Player $player) {
+        $pk = new AddPlayerPacket();
+        $pk->eid = $this->getId();
+        $pk->uuid = UUID::fromRandom();
+        $pk->type = self::NETWORK_ID;
+        $pk->x = $this->x;
+        $pk->y = $this->y;
+        $pk->z = $this->z;
+        $pk->speedX = 0;
+        $pk->speedY = 0;
+        $pk->speedZ = 0;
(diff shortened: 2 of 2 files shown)


jjplaying:
@dtkapps The old way is to spawn it as a particle, and with my PR, you spawn it as an entity. If you change the original one, you'd still need to spawn it as particle, and then, you would loose the advantages of the entity like easy despawning. If I understood you correctly ;).

src/pocketmine/entity/FloatingText.php
```diff
@@ -0,0 +1,56 @@
...
+
+namespace pocketmine\entity;
+
+use pocketmine\item\Item as ItemItem;
+use pocketmine\network\protocol\AddPlayerPacket;
+use pocketmine\Player;
+use pocketmine\utils\UUID;
+
+class FloatingText extends Entity {
+    const NETWORK_ID = 80;
```
jjplaying: I know. You need to set a network ID for an entity, otherwise, the server won't register it in the known entities list. As I don't use the ID in the code, it's completely meaningless, just a number higher than -1 needed for the server.

dktapps: That's a bad idea. It will cause creating an createEntity() with 80 to create a FTP instead of an arrow.

KairusDarkSeeker:
use 63 for the Player's network id

JackNoordhuis:
The floating text particle is a 'hack' more than anything, only actual entities with their own network ID's should be handle by the server like an entity.

src/pocketmine/entity/FloatingText.php
jjplaying: Okay, then the only way is to use an randomunsed number.

jjplaying:
@JackNoordhuis See the issue #1986

JackNoordhuis:
The floating text entity is a human entity is it not? If you're creating a human entity and using it on the server does it not seem logical to extend the Human class?

jjplaying:
@JackNoordhuis It doesn't because the Human class just adds things like food, XP, and inventory. Things which would be unnecessary for a floating text.

Tarik02:
This is untested. I'm see errors without testing!
```
	$pk = new AddPlayerPacket();
...
	$pk->type = self::NETWORK_ID;
```
And so this not fixes arrow affecting. The best way to do this, edit FTP or use this bad code in your plugin.

jjplaying:
@Tarik02 Before calling it bad code, just make a better one and read correctly. It does fix the arrow affecting, but there's a logical error (ID sent to client), I'll change it soon. And it looks like you didn't read this thread, I've already wrote why to don't edit FTP.

jjplaying:
I did some changes (0f2bc40b4c893631b485ef8f4f7c8ecc100416eb) now which fixes everything. I know, it's still a hack, but I think it's a better hack than registering it as a particle.

- NETWORK_ID in the FloatingText class is 1000, so there are no collisions with other entitiy IDs on the server.
- To the client, I'm sending AddPlayerPacket with Arrow::NETWORK_ID (80), because that's a valid item ID and is working without any problems. If you have a better idea what ID i can use, just say it to me.

I tested the FloatingText and it's working now. Sorry for a8d28fe67f8a16bcae47d9afe2cabe145b0a3e14, it was from my mobile and i didn't noticed the NETWORK_ID in spawnTo().

Nevertheless, it would be nice if some people could test the latest commit, if it works for everyone now.

jojoe77777:
@jjplaying AddPlayerPacket doesn't have a type.

JackNoordhuis:
@jjplaying You can easily override functions to remove the 'useless' stuff.

jjplaying:
@jojoe77777 Okay, didn't knew that, then I can remove it.

@JackNoordhuis No, because it's invisible. But because it need no type, it's just ignored.

JackNoordhuis:
There's a reason we're forced to use human entities for this hacky particle. You used to be able to set the custom name of any entity and it'd be visible like a player's name tag but Mojang decided to only make the name tags appear when you look at or hover over an entity that isn't a human.

My point being that if an arrow entity was actually spawned it wouldn't work thanks to Mojangs shitty client.

src/pocketmine/Server.php
```diff
@@ -2935,7 +2936,8 @@ private function registerEntities(){
 		Entity::registerEntity(Egg::class);
 		Entity::registerEntity(Enderman::class);
 		Entity::registerEntity(FallingSand::class);
-		Entity::registerEntity(FishingHook::class);
+        Entity::registerEntity(FishingHook::class);
+        Entity::registerEntity(FloatingText::class);
```
dktapps: This formatting is still incorrect. Use tabs, not spaces.

src/pocketmine/entity/FloatingText.php
```diff
@@ -0,0 +1,56 @@
...
+
+    public function setText($text) {
+        $this->text = $text;
+    }
+
+    public function spawnTo(Player $player) {
+        $pk = new AddPlayerPacket();
+        $pk->eid = $this->getId();
+        $pk->uuid = UUID::fromRandom();
+        $pk->type = Arrow::NETWORK_ID;
```
dktapps: This is unnecessary.

dktapps:
My main objection to this is that there are now _two_ implementations of FloatingTextParticle. This will lead to confusion between the two. This does not really fix the issue with FTP despawning as plugins will still have to be changed to use this new implementation. I think it's doable to make the old implementation use this.

src/pocketmine/Server.php
jjplaying: I'm using PhpStorm, and just create a new line, and it is the right distance. Even if you look it in GitHub, the distance is right. Blame PhpStorm if it's wrong.

jjplaying:
@dktapps I can do a way with no change for plugins, so I'll use @legoboy0215's hack from here (https://github.com/iTXTech/Genisys/issues/1986). But they won't despawn automatically, you need to send a RemoveEntityPacket on player quit in Genisys specially for FloatingTexts.

The two possibilities:
- You can do the old particle way, maybe fix it and make it better, there will be no changes for plugins, but it will always be a hack and you need to despawn it manually (in Genisys code on PlayerQuitEvent)

- Or you use the new entity way, it's not so hacky like particle, but then, all plugins need a change (or you keep both, but then, it's confusing).

All ways have disadvantages, because MCPE simply does not know FloatingText's officially, they will be always hacky.

It's just the question which hack we should use. I'm working on both at this time (https://github.com/jjplaying/Genisys/tree/floating-as-entity / https://github.com/jjplaying/Genisys/tree/floating-as-particle). Maybe dtkapps should decide or we should vote for a way.

Tarik02:
@jjplaying, 
"All ways have disadvantages, because **MCPE simply does not know FloatingText's officially**", - it will be, believe.

jjplaying:
@Tarik02 Hoping on it ;)

jjplaying:
Sorry, I deleted the branch by mistake, but restored it.

Misteboss:
So finally it wil be added ? @dktapps

dktapps:
If you continue to spam at me you will be banned. That includes spamming at me on Twitter.

This is not my full-time job, and this PR is not yet ready to be merged.

XFuryMCPE:
this pull needs to be updated to work with 0.16 flag changes

KairusDarkSeeker:
Simple solution: $level->addParticle(FloatingTextParticle, $level->getPlayers());

XFuryMCPE:
That isn't a solution to the 0.16 flag changes ..

KairusDarkSeeker:
@XFuryMCPE That was a solution for making FTPs on all worlds

ishitatsuyuki:
Since there's merge conflict and this has been controversial, I'm closing it; feel free to request for a reopen if you have arguments.

PR #2054, opened 2016-10-10
Fix #2030

### LINKED ISSUE #2030: kill() in ThrownPotion runs twice when it hit the Player that thrown it
#### Issue description

kill() in ThrownPotion runs twice when it hit the Player that thrown it.

Add a echo; to the kill(), its going to run twice which triggers the parent::kill() which is twice set the $this->health = 0. Isn't that means nothing ?

#### OS and versions

* Genisys: 498379b
* PHP: 7.3
* Server OS: Windows 10
* Game version: W10E

```
public function kill(){
	    echo 123;
		$color = Potion::getColor($this->getPotionId());
		$this->getLevel()->addParticle(new SpellParticle($this, $color[0], $color[1], $color[2]));
		$players = $this->getViewers();
		foreach($players as $p) {
			if($p->distance($this) <= 6){
				foreach(Potion::getEffectsById($this->getPotionId()) as $effect){
					$p->addEffect($effect);
				}
			}
		}
		
		parent::kill();
	}
```

JackNoordhuis:
### Description
Fixes multiple calls to ThrownPotion::kill() and prevents potion effects being applied multiple times in the event a plugin attempts to kill the entity as well.

### Reason to modify
Fixes #2030 and prevents plugins from causing the same issue.

### Tests & Reviews
I've tested the changes on my personal test server thoroughly and everything works.

### DIFF
--- a/src/pocketmine/entity/ThrownPotion.php
+++ b/src/pocketmine/entity/ThrownPotion.php
@@ -51,24 +51,26 @@ class ThrownPotion extends Projectile{
 		unset($this->dataProperties[self::DATA_SHOOTER_ID]);
 		$this->setDataProperty(self::DATA_POTION_ID, self::DATA_TYPE_SHORT, $this->getPotionId());
 	}
-	
+
 	public function getPotionId() : int{
 		return (int) $this->namedtag["PotionId"];
 	}
-	
+
 	public function kill(){
-		$color = Potion::getColor($this->getPotionId());
-		$this->getLevel()->addParticle(new SpellParticle($this, $color[0], $color[1], $color[2]));
-		$players = $this->getViewers();
-		foreach($players as $p) {
-			if($p->distance($this) <= 6){
-				foreach(Potion::getEffectsById($this->getPotionId()) as $effect){
-					$p->addEffect($effect);
+		if($this->isAlive()) {
+			$color = Potion::getColor($this->getPotionId());
+			$this->getLevel()->addParticle(new SpellParticle($this, $color[0], $color[1], $color[2]));
+			$players = $this->getViewers();
+			foreach($players as $p) {
+				if($p->distance($this) <= 6) {
+					foreach(Potion::getEffectsById($this->getPotionId()) as $effect) {
+						$p->addEffect($effect);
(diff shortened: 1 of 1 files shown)


Tarik02:
All ok, but this commit not must be here.
https://github.com/iTXTech/Genisys/pull/2054/commits/5271e7a1c8e2bfcd4930e869947a066ff4883711

JackNoordhuis:
Squashed commits, didn't have time todo it yesterday.

Tarik02:
All ok, tested.

src/pocketmine/tile/Hopper.php
```diff
@@ -99,7 +99,7 @@ public function onUpdate(){
 		$area = clone $this->getBlock()->getBoundingBox(); //Area above hopper to draw items from
 		$area->maxY = ceil($area->maxY) + 1; //Account for full block above, not just 1 + 5/8
 		foreach($this->getLevel()->getChunkEntities($this->getBlock()->x >> 4, $this->getBlock()->z >> 4) as $entity){
-			if(!($entity instanceof DroppedItem)){
+			if(!($entity instanceof DroppedItem) or !$entity->isAlive()){
```
dktapps: Seems an extra change sneaked in here from your last PR. Please correct this.

src/pocketmine/entity/ThrownPotion.php
```diff
@@ -86,11 +88,10 @@ public function onUpdate($currentTick){
 			$this->kill();
 			$this->close();
 			$hasUpdate = true;
-		}
-		
-		if($this->onGround) {
+		} elseif($this->onGround) {
 			$this->kill();
```
dktapps: This condition seems somewhat useless. Why not put $this->onGround check in the condition above?

JackNoordhuis: Wasn't looking to mess with the logic, just trying to fix the bug. I'll do some tests and see if that condition is needed.

JackNoordhuis:
The onGround check never occurs due to isCollided also being returning true if the entity has come in contact with a block.

Issue #2057, opened 2016-10-11
My server connects is not giving error on the console


MarceloJr21:
```
Genisys Crash Dump Tue Oct 11 05:14:35 EDT 2016

Error: Argument 1 passed to pocketmine\permission\BanEntry::setCreated() must be an instance of DateTime, boolean given, called in phar:///home/multicraft/jar/Genisys.phar/src/pocketmine/permission/BanEntry.php on line 112
File: /src/pocketmine/permission/BanEntry
Line: 48
Type: notice

Code:
[39] 
[40] 	public function getName() : string{
[41] 		return $this->name;
[42] 	}
[43] 
[44] 	public function getCreated(){
[45] 		return $this->creationDate;
[46] 	}
[47] 
[48] 	public function setCreated(\DateTime $date){
[49] 		$this->creationDate = $date;
[50] 	}
[51] 
[52] 	public function getSource(){
[53] 		return $this->source;
[54] 	}
[55] 
[56] 	public function setSource($source){
[57] 		$this->source = $source;
[58] 	}

Backtrace:
#0 /src/pocketmine/permission/BanList(137): pocketmine\permission\BanEntry::fromString(array Array())
#1 /src/pocketmine/Server(1849): pocketmine\permission\BanList->load(boolean)
#2 /src/pocketmine/PocketMine(476): pocketmine\Server->__construct(pocketmine\CompatibleClassLoader object, pocketmine\utils\MainLogger object, string phar:///home/multicraft/jar/Genisys.phar/, string /home/multicraft/servers/server7346/, string /home/multicraft/servers/server7346/plugins/, string unknown)
#3 (1): require_once(string phar:///home/multicraft/jar/Genisys.phar/src/pocketmine/PocketMine.php)

Genisys version: b816e46335cf145b8eca9c1a24b18ee10e80cae3 [Protocol 84; API 2.0.0]
uname -a: Linux s6.craftandohosting.net 3.14.32-xxxx-grs-ipv6-64 #7 SMP Wed Jan 27 18:05:09 CET 2016 x86_64
PHP version: 7.0.10
Zend version: 3.0.0
OS : Linux, linux

Server uptime: 0 seconds
Number of loaded worlds: 0
Players online: 0/
```

FaigerSYS:
Delete/fix/remove last entry in your ban list (banned-[somesing].txt)

FaigerSYS:
[...]

PR #2059, opened 2016-10-11
Ban fix

FaigerSYS:
### Description

This PR prevents #2057 issue


src/pocketmine/lang/locale/eng.ini
```diff
@@ -77,6 +77,7 @@ commands.generic.exception=An unknown error occurred while attempting to perform
 commands.generic.permission=You do not have permission to use this command
 commands.generic.notFound=Unknown command. Try /help for a list of commands
 commands.generic.player.notFound=That player cannot be found
+commands.generic.player.invalidName=Invalid name of player
```
jojoe77777: Why not just commands.generic.player.notFound, above it?

src/pocketmine/command/defaults/BanCommand.php
```diff
@@ -50,6 +50,18 @@ public function execute(CommandSender $sender, $currentAlias, array $args){
 		}
 
 		$name = array_shift($args);
+		$len = strlen($name);
+		for($i = 0; $i < $len; ++$i){
+			$c = ord($name{$i});
+			if(($c >= ord("a") and $c <= ord("z")) or ($c >= ord("A") and $c <= ord("Z")) or ($c >= ord("0") and $c <= ord("9")) or $c === ord("_")){
```
iPocket: Why? use regex.

dktapps: because this code was copy-pasted...

src/pocketmine/command/defaults/BanCommand.php
```diff
@@ -50,6 +50,18 @@ public function execute(CommandSender $sender, $currentAlias, array $args){
...
+				continue;
+			}
```
SOF3: Can be simplified into using `ctype_alnum(str_replace("_", "", $name))`.

FaigerSYS: @SOF3 , I don't know PHP so good like u :P

PR #2061, opened 2016-10-11
Fixes #2057

### LINKED ISSUE #2057: My server connects is not giving error on the console

jjplaying:
### Description

Fixes #2057, it replaces | with nothing to avoid bugs in the ban.txt.

### Tests & Reviews

Tested the commands and it works.

### DIFF
--- a/src/pocketmine/command/defaults/BanCommand.php
+++ b/src/pocketmine/command/defaults/BanCommand.php
@@ -49,9 +49,9 @@ class BanCommand extends VanillaCommand{
 			return false;
 		}
 
-		$name = array_shift($args);
+		$name = str_replace("|", "", array_shift($args));
 		if(isset($args[0]) and isset($args[1])){
-			$reason = $args[0];
+			$reason = str_replace("|", "", $args[0]);
 			if($args[1] != null and is_numeric($args[1])){
 				$until = new \DateTime('@' . ($args[1] * 86400 + time()));
 			}else{
--- a/src/pocketmine/command/defaults/BanCommand.php
+++ b/src/pocketmine/command/defaults/BanCommand.php
@@ -49,9 +49,9 @@ class BanCommand extends VanillaCommand{
 			return false;
 		}
 
-		$name = str_replace(array("|", "_"), "", array_shift($args));
+		$name = ctype_alnum(str_replace("_", "", array_shift($args)));
 		if(isset($args[0]) and isset($args[1])){
-			$reason = str_replace(array("|", "_"), "", $args[0]);
+			$reason = ctype_alnum(str_replace("_", "", $args[0]));
 			if($args[1] != null and is_numeric($args[1])){
 				$until = new \DateTime('@' . ($args[1] * 86400 + time()));
 			}else{
--- a/src/pocketmine/command/defaults/BanCommand.php
+++ b/src/pocketmine/command/defaults/BanCommand.php
@@ -49,9 +49,14 @@ class BanCommand extends VanillaCommand{
 			return false;
 		}
 
-		$name = ctype_alnum(str_replace("_", "", array_shift($args)));
+		$name = str_replace("_", "", array_shift($args));
 		if(isset($args[0]) and isset($args[1])){
-			$reason = ctype_alnum(str_replace("_", "", $args[0]));
+			$reason = str_replace("_", "", $args[0]);
+			
+			if(!ctype_alnum($name) OR !ctype_alnum($reason) {
+				return false;
+			}
+			
 			if($args[1] != null and is_numeric($args[1])){
 				$until = new \DateTime('@' . ($args[1] * 86400 + time()));
 			}else{
--- a/src/pocketmine/command/defaults/BanCommand.php
+++ b/src/pocketmine/command/defaults/BanCommand.php
@@ -49,9 +49,9 @@ class BanCommand extends VanillaCommand{
 			return false;
 		}
 
-		$name = str_replace("|", "", array_shift($args));
+		$name = str_replace(array("|", "_"), "", array_shift($args));
 		if(isset($args[0]) and isset($args[1])){
-			$reason = str_replace("|", "", $args[0]);
+			$reason = str_replace(array("|", "_"), "", $args[0]);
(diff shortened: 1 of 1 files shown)


dktapps:
That isn't the only character that causes such a crash.

jjplaying:
Which ones? Someone sad it also crashes with `_`.

dktapps:
This isn't really an effective solution. You should whitelist allowed characters, not blacklist disallowed ones. There are myriad characters this could occur wity.

FaigerSYS:
@jjplaying , false. "_" symbol didn't crash server. Revert this

FaigerSYS:
As @SOF3 said, you can just use `ctype_alnum(str_replace("_", "", $name))` check

jjplaying:
Sounds logical, I'll change this.

jjplaying:
Oh, didn't checked this, too.

jjplaying:
return false; should be okay, because nobody can have those characters in his name.

jjplaying:
Oh shit I can't write on my phone, fix coming soon.

src/pocketmine/command/defaults/BanCommand.php
```diff
@@ -49,9 +49,14 @@ public function execute(CommandSender $sender, $currentAlias, array $args){
 			return false;
 		}
 
-		$name = ctype_alnum(str_replace("_", "", array_shift($args)));
+		$name = str_replace("_", "", array_shift($args));
```
FaigerSYS: You shouldn't replace "_" symbol. You can just remove str_replace here.
`$name = array_shift($args);
if (!ctype_alnum(str_replace("_", "", $name))) {
    $sender->sendMessage("Error message");
    return false;
}`

src/pocketmine/command/defaults/BanCommand.php
```diff
@@ -49,9 +49,14 @@ public function execute(CommandSender $sender, $currentAlias, array $args){
...
 		if(isset($args[0]) and isset($args[1])){
-			$reason = ctype_alnum(str_replace("_", "", $args[0]));
+			$reason = str_replace("_", "", $args[0]);
```
FaigerSYS: There you can just replace this symbol. Or use another filter
`$reason = str_replace("|", "", $args[0]);`
The same must be used (there)[https://github.com/iTXTech/Genisys/blob/master/src/pocketmine/command/defaults/BanCommand.php#L63]

src/pocketmine/command/defaults/BanCommand.php
dktapps: @FaigerSYS Go and read the PHP documentation instead of spamming incorrect comments.

jjplaying:
I'll rewrite it tomorrow.

dktapps:
I think ban lists as YAML or some other config format would make more sense, then there's no restriction on what you can and can't do.

jjplaying:
Yeah, that's a very good idea. I'll start writing such a thing today evening when I'm back from school.

FaigerSYS:
@jjplaying , #2073

jjplaying:
Closed

FaigerSYS:
@jjplaying , new PR. #2078

jjplaying:
Looks good, but also closed. We should dtkapps let do this, as he sad, it was his idea.

You should use nested YAML and not the old format in YAML files, that's better.

ishitatsuyuki:
Some principles:
Keep it simple, stupid
You're not gonna need it

So, only fixing the bug should be enough. Anyway, I'm gonna reviewing it soon.

SOF3:
The output name can contain an underscore...

ishitatsuyuki:
@dktapps if you are free, please give a review.

dktapps:
This doesn't actually fix the initial issue, which is that an invalid banned-players.txt would cause a crash. I fixed that in https://github.com/iTXTech/Genisys/commit/553f2a5a9e75ff2aa1bc3d7d7096323b3f71cc35

This also isn't foolproof. The main issue here is the separator |. Potential issues can also arise from putting a | in the reason, for example.

Finally, I think such a name validation should be a utils static function for consistency throughout the software.

Issue #1962, opened 2016-09-20
Some bugs


GenisysFTW:
#### Issue description

Bugs with armor and absorption effect.
Armor: Armor takes no damage.
Absorption effect: This effect is glitchy. It'll take your main hearts for a second and replace it with absorption hearts so you have 18 health + 4 absorption health.

#### Steps to reproduce the issue

Armor damage: Equip an armor and try calling, damaging yourself by lava or cactus.
Absorption: Give yourself absorption and try damaging yourself by cactus or lava. Keep an eye on the hearts and see your last main heart getting replaced by a golden heart.

#### OS and versions

* Genisys: 5e4bfc0
* PHP: 7.0.10
* Server OS: Ubuntu 14.04 LTS (64-bit)
* Game version: 0.15.8

#### Crashdump, backtrace, memory dumps, plugins list or other files

```
No crash dumps.
No console errors.
No traces.
```

SOF3:
There are two bugs in this issue?

GenisysFTW:
I have a feeling these two bugs are related.

ImagicalGamer:
I can confirm the Armor bug.

corytortoise:
I can also confirm the armor bug. Has anybody found a workaround?

IrishUHC:
I can confirm the absorption bug.

dktapps:
The absorption bug is a client-sided issue, it's also reproducible in vanilla. It appears to occur when the player has an odd number of absorption points.

PR #2062, opened 2016-10-11
FIX no Armor Damage - #1962

laurobeleche:
### Description
To fix the bug of no Armor damage

### Reason to modify
I tested within my server initially created a plugin that corrected the bug, then I realized that the function that creates the damage the armor was not being called, then traced the causes that damage the armor and put to call the function.

Fix no Armor damage

### DIFF
--- a/src/pocketmine/event/entity/EntityDamageEvent.php
+++ b/src/pocketmine/event/entity/EntityDamageEvent.php
@@ -159,7 +159,6 @@ class EntityDamageEvent extends EntityEvent implements Cancellable{
 							}
 						}
 					}
-					$this->useArmors();
 					break;
 				case self::CAUSE_FALL:
 					//Feather Falling
--- a/src/pocketmine/event/entity/EntityDamageEvent.php
+++ b/src/pocketmine/event/entity/EntityDamageEvent.php
@@ -159,6 +159,7 @@ class EntityDamageEvent extends EntityEvent implements Cancellable{
 							}
 						}
 					}
+					$this->useArmors();
 					break;
 				case self::CAUSE_FALL:
 					//Feather Falling
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -4020,6 +4020,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 				$this->exhaust(0.3, PlayerExhaustEvent::CAUSE_DAMAGE);
 			}
 		}
+		return true;
 	}
 
 	public function sendPosition(Vector3 $pos, $yaw = null, $pitch = null, $mode = 0, array $targets = null){
--- a/src/pocketmine/inventory/CraftingManager.php
+++ b/src/pocketmine/inventory/CraftingManager.php
@@ -187,6 +187,11 @@ class CraftingManager{
 				"PSP",
 				"   "
 			))->setIngredient("S", Item::get(Item::STICK, 0, 2))->setIngredient("P", Item::get(Item::WOODEN_PLANK, Planks::OAK, 4)));
+			$this->registerRecipe((new BigShapedRecipe(Item::get(Item::NETHER_BRICK_FENCE, 0, 3),
+ 				"PPP",
+ 				"PPP",
+ 				"   "
+ 			))->setIngredient("P", Item::get(Item::NETHER_BRICKS, 0, 6)));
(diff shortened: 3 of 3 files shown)


src/pocketmine/event/entity/EntityDamageEvent.php
```diff
@@ -159,6 +159,7 @@ public function __construct(Entity $entity, $cause, $damage){
 							}
 						}
 					}
+					$this->useArmors();
```
dktapps: This is one reason I tell people not to use the web editor. That method does not exist in EntityDamageEvent nor any of its parents.

laurobeleche: But then what editor I use, I have it on my PC and edit with Notepad ++, this change fixed the bug.

laurobeleche:
`/**
	 * @return bool
	 */
	public function useArmors(){
		if($this->entity instanceof Player){
			if($this->entity->isSurvival() and $this->entity->isAlive()){
				foreach($this->usedArmors as $index => $cost){
					$i = $this->entity->getInventory()->getArmorItem($index);
					if($i->isArmor()){
						/** @var Armor $i */
						$i->useOn($i, $cost);
						$this->entity->getInventory()->setArmorItem($index, $i);
					}
				}
			}
			return true;
		}
		return false;
	}`
This function is rather the event on line 301 you can check the method, and it works, just be called as I did in my correction.

dktapps:
>_> blame me for using a mobile device to review code. My bad, sorry.

src/pocketmine/event/entity/EntityDamageEvent.php
dterweij: PHPStorm IDE.

dog194:
no no no... useArmors() will reduce the damage which is list in $this->usedArmors.
and you should not use this function in event!!!!!!!!!
this function should only trigger when the player really get hurt. 

e.g. if a player walk into a Fire block. This event will be called more than 10 times in 1 second. With player have damage cooldown. so just one of this event will really hurt the player.!!!
reference: https://github.com/iTXTech/Genisys/search?utf8=%E2%9C%93&q=useArmors%28%29

BTW i am really busy recent days. QAQ so for this

dog194:
fewer months ago, when this event change. Armors Durability System work fine with walking into fire lava and Cactus. So maybe you should have a check useArmors() function. 

I'm sorry that I can not help due to a  ton of my daily work.

laurobeleche:
but then why the function has not been called in the case of PVP damage?

dog194:
@laurobeleche     
src/pocketmine/Player.php   line 3013

laurobeleche:
But the function has not been executed, the bug's there a long time and NOBODY is doing nothing to fix it. I mobilized, I opened the door, showed the way, I have little experience with pocketmine / Genisys, but I will delve to remedy the various bugs that harm us every day.

laurobeleche:
In my tests, I identified the
if ($ target-> attack (EV $> getFinalDamage (), $ ev) === true)
There was returning TRUE in any way, then I added a true return to function so that the condition was true.
Thanks @dog194 @dktapps

SuperDude36:
I downloaded Genisys on Jenkins and armor still doesn't break.

laurobeleche:
https://gitlab.com/laurobeleche/Genisys/builds/5078359/artifacts/download

Download this,

SuperDude36:
Eh, 404 error.

laurobeleche:
https://gitlab.com/laurobeleche/Genisys/pipelines

Commit players.php, download artfacts

SuperDude36:
Still 404 error...

laurobeleche:
Sorry

http://www.mediafire.com/file/zal9m7xu19m1zdr/Genisys_bef53a6_2016-10-13_02-18-58.phar

dog194:
errrrr. For some reason i cannot help you to have a check now. 
So maybe merge until getting more than 3 test result from other user~.
Note:   PVP  walk into fire block and so on
need to check the armor only lost 1 durability when get hurt.

SuperDude36:
Thanks for the media fire link!!! Works!

ishitatsuyuki:
You created redundant conflicts. Learn to use git.

laurobeleche:
Sorry

ishitatsuyuki:
**Use Git or close.**

laurobeleche:
Sorry, I installed git and now I started working with him.

PR #2069, opened 2016-10-13
Allow cancelling hunger healing

XFuryMCPE:
### Description

You can now cancel EntityRegainHealthEvent when the cause is by hunger

### Reason to modify

^^^^

### Tests & Reviews


IrishUHC:
Good PR, I have no idea why it wasn't in there in the first place.

dktapps:
Perhaps because of this? https://github.com/iTXTech/Genisys/blob/master/src/pocketmine/entity/Living.php#L91

dktapps:
Correction: _this_
https://github.com/iTXTech/Genisys/blob/master/src/pocketmine/entity/Entity.php#L703

dktapps:
> - Have you made sure that there is actually a problem before trying to fix it?

Please learn, or I'm going to ban you again.

PR #2075, opened 2016-10-13
Update craftings

Gianluxx:
### Description
Update Craftings

### Reason to modify
^^^^

### Tests & Reviews

Added some craftings, removed duplicates (wool was written 4 times->deleted 3), fixed some

### DIFF
--- a/src/pocketmine/inventory/CraftingManager.php
+++ b/src/pocketmine/inventory/CraftingManager.php
@@ -188,10 +188,10 @@ class CraftingManager{
 				"   "
 			))->setIngredient("S", Item::get(Item::STICK, 0, 2))->setIngredient("P", Item::get(Item::WOODEN_PLANK, Planks::OAK, 4)));
 			$this->registerRecipe((new BigShapedRecipe(Item::get(Item::NETHER_BRICK_FENCE, 0, 3),
-				"PSP",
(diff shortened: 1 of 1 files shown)


src/pocketmine/inventory/CraftingManager.php
```diff
@@ -419,8 +424,8 @@ public function __construct(bool $useJson = false){
 			$this->registerRecipe((new BigShapedRecipe(Item::get(Item::BOOK, 0, 1),
 				"   ",
 				"PP ",
-				"P  "
-			))->setIngredient("P", Item::get(Item::PAPER, 0, 3)));
+				"PL "
```
dktapps: This is incorrect. The recipe you've given is the PC recipe.

Gianluxx: is correct.. i wrote by myself and works

Gianluxx: why can you create a book without leather? it's stupid

dktapps: That's the PC recipe  This is a PE server.

dktapps: Hmm, they changed it in a recent update...

PR #2078, opened 2016-10-14
YAML ban system

FaigerSYS:
Add YAML ban system. Implemented switcher and updater from default to YAML


ishitatsuyuki:
Please review the contribution guidelines before opening a pull request.

[Contribution guidelines](https://github.com/iTXTech/Genisys/blob/master/CONTRIBUTING.md)
> **Do not consider spamming pull requests.**

FaigerSYS:
@ishitatsuyuki , ummm, I closed previous PR because of it was "dirty". I rewrited YAML system, and published it there. So please reopen this PR

SOF3:
Multiple ban list formats existing at the same time is definitely a bad idea. Deprecate the old one instead of making an option.

FaigerSYS:
@SOF3 , it's not multiple, only one works at the same time

SOF3:
@FaigerSYS it is still multiple in the same version. It is a bad idea to support multiple formats of a config file in the same version, as it can easily to confusion between the two in the future, and there being no automatic interconversion.

FaigerSYS:
@SOF3 , I would like to rewrite, but PR is closed

src/pocketmine/permission/BanList.php
```diff
@@ -155,13 +152,41 @@ public function save($flag = true){
...
 			}
+
+			fwrite($fp, yaml_emit($data));
 			fclose($fp);
 		}else{
 			MainLogger::getLogger()->error("Could not save ban list");
 		}
 	}
 
+	public static function __toYaml($old_file, $new_file) {
```
SOF3: Avoid naming methods starting with `__`. They may clash with PHP magic methods.

src/pocketmine/permission/BanList.php
```diff
@@ -37,6 +37,7 @@ class BanList{
 
 	/**
 	 * @param string $file
+	 * @param bool $isYaml
```
SOF3: This line can be removed.

src/pocketmine/permission/BanEntry.php
```diff
@@ -127,4 +122,5 @@ public static function fromString($str){
 			return $entry;
 		}
 	}
+
```
SOF3: Unnecessary change

src/pocketmine/permission/BanEntry.php
```diff
@@ -82,43 +82,38 @@ public function setReason($reason){
...
+		return $data;
 	}
 
 	/**
-	 * @param string $str
+	 * @param array $data
 	 *
 	 * @return BanEntry
 	 */
-	public static function fromString($str){
```
SOF3: This is not the proper way to deprecate a method. You should remain basic support for the method, and mark it as deprecated.

src/pocketmine/permission/BanList.php
dktapps: The __ is reserved for use with magic methods. This is not a magic method.

src/pocketmine/permission/BanEntry.php
```diff
@@ -82,43 +82,38 @@ public function setReason($reason){
...
-		if(strlen($str) < 2){
+	public static function fromArray(array $data){
+		if(count($data) < 1){
```
SOF3: why not an empty BanList?

src/pocketmine/permission/BanEntry.php
```diff
@@ -82,43 +82,38 @@ public function setReason($reason){
 		$this->reason = $reason;
 	}
 
-	public function getString(){
```
SOF3: This method should be marked as deprecated instead of being removed.

dktapps:
This pull request does not make effective use of the YAML format. It produces a format like:
```
---
- - dktapps
  - "2016-10-14 17:14:47 +0800"
  - CONSOLE
  - Forever
  - forever
...
```
This is a very inefficient way to store such data, and does not utilise the useful features that YAML provides.

FaigerSYS:
@dktapps , what do you suggest? The main thing that there were no errors

SOF3:
@FaigerSYS if we just want no errors, don't use YAML. Just serialize all the stuff and that gets done.

We want a ban list that is easy to read and edit for both programs and users.

dktapps:
I think I'll come up with something myself. This is far too much palaver for something like this, and it _was_ my idea to begin with.

FaigerSYS:
@SOF3 , so you say that format is not easy to read and edit by users.....?

PR #2088, opened 2016-10-16
Crafting update

inferboy:
### Description

Book Crafting Update!

### Reason to modify

Better leather support to the book so the pages are binded properly and doesn't tear up because we all hate when it starts to tear up... Quality matters.
### Tests & Reviews


dktapps:
This is not the vanilla recipe. Closing.

inferboy:
But Quality matters

PR #2114, opened 2016-10-22
Fixed genisys_eng.yml language

IISweePII:
### Description
This pull request is for for simplifying the explanation for the configuration file settings for Genisys, I have bumped the version # to 22 now (before it was version 21), since this is a change to the config and bumping it up 1 number would make it seem better for it to be applied to.

### Reason to modify
The reason I decided to do this change to the file is because simply I am good at English speaking and grammar and I can fix any errors that are present in grammar, also these lines I edited were comments, so it would have no effect to the game having any errors (no php code change (only comment change)).

### Tests & Reviews
I have now tested this code with the travis.ci thing, looking at the changes that were done, there was only comments to help explain what each setting did, so it would have no effect to the game itself (no src error or anything). I have put a good amount of thought into how to sound each comment out, I did allow edits from the maintainers of this projects if something seems to be a issue. 

This is my first time making a PR and contributing to this project, so I am new to the whole process, thank you once again for looking through this. - Sweep

### DIFF
--- a/src/pocketmine/resources/genisys_eng.yml
+++ b/src/pocketmine/resources/genisys_eng.yml
@@ -2,7 +2,7 @@
 
 #Version of this file
 config:
- version: 22
+ version: 21
 
 level:
  #Set if weather is enabled (rain may cause lag to older devices)
--- a/src/pocketmine/resources/genisys_eng.yml
+++ b/src/pocketmine/resources/genisys_eng.yml
@@ -2,37 +2,37 @@
 
 #Version of this file
 config:
- version: 21
+ version: 22
 
 level:
- #Set if weather is allowed
+ #Set if weather is enabled (rain may cause lag to older devices)
  weather: true
  #Weather random duration
  weather-random-duration-min: 6000
  weather-random-duration-max: 12000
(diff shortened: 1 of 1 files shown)


src/pocketmine/resources/genisys_eng.yml
```diff
@@ -2,37 +2,37 @@
 
 #Version of this file
 config:
- version: 21
+ version: 22
```
dktapps: This is unnecessary, no property changes have been made, only comments.

IISweePII: Thank you for allowing that, I thought I had to change the version #, I made it easier for it to understand. I will change it! :)

IISweePII:
@dktapps I have fixed the version # to 21 now, I have re-ran the commit, it is working as well. Should I make a new PR for it?

TheAppleGamer:
Seems good!

IISweePII:
@TheAppleGamer Fixing things like this is what I feel better towards, thanks for the feedback haha, I took some time thinking of what to put in each area. At least Dylan was alright with it, just the Version # change had no need lol

PR #2122, opened 2016-10-24
FIX no Armor Damage - #1962

laurobeleche:
Description

To fix the bug of no Armor damage

Reason to modify

I tested within my server initially created a plugin that corrected the bug, then I realized that the function that creates the damage the armor was not being called, then traced the causes that damage the armor and put to call the function.

Fix no Armor damage

Creating again with git use


ishitatsuyuki:
1. You're spamming pull request.
2. Squash.

PR #2128, opened 2016-10-26
Tile coordinates

NoHaxJustTryHqrd:
### Description

Easily get a tiles coordinates

### Reason to modify
 it will be easier to get a tiles coords

### Tests & Reviews

Yes it should infact work


dktapps:
This is totally useless, since Tile extends Position, which is an extension of Vector3, which already has these methods.

NoHaxJustTryHqrd:
My bad for not seeing that

PR #2133, opened 2016-10-30
Fixes https://github.com/iTXTech/Nemisys/issues/80 in 0.16 branch

jjplaying:
### Description

Fixes https://github.com/iTXTech/Nemisys/issues/80 in 0.16 branch

In the genisys.yml, the synapse server password is named "server-password", and in the code, just "password" is used, so this fixes the bug.


jjplaying:
Oh thats completely wrong I messed up my development things, sorry.

PR #2138, opened 2016-11-03
Remove synapse

Tarik02:
### Description

This PR removes synapse from genisys.

### Reason to modify

The better way to use synapse is such as SynapseAPI fro Nukkit. I'm created this [plugin](https://github.com/Tarik02/SynapsePM).

### Tests & Reviews


SleepSpace9:
Great! This is exactly the way as I imagined it. :heart: 
My own plugins which uses SynapseAPI methods will still work and the bloating of Genisys codebase went out. :+1:

dktapps:
Thank you so much, I've wanted to do this for a very long time. Synapse is just added unneeded bloat for me, and yet I didn't want to risk removing.

I wonder, will the plugin you mentioned work on PocketMine itself, or is it Genisys-only?

vvzar:
Congratulations! It`s awesome idea!

Tarik02:
@dktapps, this plugin must works also with PocketMine, but i will test it soon.

Tarik02:
@dktapps, I tested this with pmmp/PocketMine-MP and it works good.

dktapps:
oops...

AryToNeX:
Great but in my opinion the plugin should be done from iTXTech, not from third-party sources and the two remaining server functions regarding Synapse on Server.php should be also deleted.

If someone wants to get a SynapseAPI instance he should get it from:
`$this->getServer()->getPlugin('SynapseAPI');`

Tarik02:
@AryToNeX, for now, methods in `Server.php` is labeled as deprecated. It will be removed. I'm recommend to use it as plugin:
```
$this->getServer()->getPlugin('SynapsePM');
```

About posting it from @iTXTech, i has no accea to publish it by @iTXTech. I want to move it to @iTXTech.

MichaelM04:
This is a great idea! If all works well, PocketMine-MP can use this and that will come in great use in the future. As for the move, I believe it would be great if iTXTech could move this to their repositories, as this would be better for future development and visibility to all.

Tarik02:
@MichaelM04, plugin works well in PocketMine-MP. About moving to iTXTech, I allow this.

Guillaume351:
@Tarik02 Do you plan  on adding SteadFast support ? When I try to start it, I get : 

> [13:38:58] [Server thread/INFO]: Enabling SynapsePM v1.0.0

Fatal error: Uncaught Exception: Serialization of 'Closure' is not allowed in [no active file]:0
Stack trace:
#0 {main}
  thrown in [no active file] on line 0

Tarik02:
@Tarik02, i will try to fix this error soon.

NycuRO:
@Guillaume351 SteadFight have proxy support -_- , use proxy branch.

Guillaume351:
@NycuRO But it's outdated, no ? Can't get it to work

NycuRO:
@Guillaume351 For me work.

jjplaying:
Your plugin seems to be untested, commands are very buggy (working well in Genisys).

Tarik02:
@jjplaying, you saying about 0.16? Plugin is not a reason for it. Try to use commands in 0.16 with Genisys but without plugin.

Muqsit:
Oww yeah! GG!

MichaelM04:
Good decision

PR #2144, opened 2016-11-04
Particle.php needs improvement.

Asidert:
### Description
In 0.16.0 devs changes numbers of some particles. (CriticalParticle have #2 again)
FlameParticle now have #7, for example.
I haven't time for test all particles. (Some now will crash client, be carefully).

### Tests & Reviews
If I call LavaParticle will spawn FlameParticle, i think you can keep this idea...


dktapps:
Already under testing, an updated list will be released when we have all the correct names.

PR #2154, opened 2016-11-05
Remove irc chat

inferboy:
### Description

Update Readme, remove useless things

### Reason to modify

IRC doesn't work and is not needed!
### Tests & Reviews

I have tested the code and it works


iPocket:
Wow, seriously

ishitatsuyuki:
You are idiot and you are not needed!

Well, just joking, IRC is somehow a lightweight and fun platform, if you could use it correctly.

PR #2158, opened 2016-11-05
Fix from Steadfast

TheAppleGamer:
### Description

Fixes the click speed for combat, it will prevent players from getting extra damage on hitting sometimes.

### Reason to modify

Players prefer the old combat system
### Tests & Reviews

I am unable to test as my test devices are not working.
If anyone could test it, that would be great!


dktapps:
Copy-paste job and untested? Bad move.

TheAppleGamer:
Copy and paste job? I did that on Notepad++ and my test devices aren't working

TheAppleGamer:
Tested

PR #2200, opened 2016-11-11
BossEventPacket from Elywing

TheAppleGamer:
### Description

Adds BossEventPacket to the protocol 

### Reason to modify

It allows players to make plugins using the bossbar
### Tests & Reviews
I've tested the code and it works
I've also received permission to use this from an Elywing member


dktapps:
That's incomplete, it's being worked on internally already.

dktapps:
Also, stuff your "permissions". PocketMine and all its variants are under the GPL open-source licence or variants of, which allows you to do what you want with said code, only stipulating that you must make your changes available under the same licence. If you need permission to copy-paste from a spoon of PocketMine, that's in violation of the licence and said spoon can as such be subjected to a DMCA claim.

PR #2203, opened 2016-11-11
Fix bad effects on Creative, close #2193

### LINKED ISSUE #2193: Potion damage in Creative
#### Issue description

When using a harming potion you can kill players in creative

Get a harming potion and a player
Make sure the player is in gmc
hit then with harming
#### OS and versions

* Genisys: 379a533
* PHP: 7.0.12
* Server OS: Windows 10
* Game version: 0.16.1

```
```

iPocket:


src/pocketmine/Player.php
```diff
@@ -4337,4 +4337,14 @@ public static function getChunkCacheFromData($chunkX, $chunkZ, $payload, $orderi
 		return $pk;
 	}
 
-}
+	/**
+	 * @param Effect
+	*/
+	public function addEffect(Effect $effect){//Overwrite
+		if($effect->isBad() && $this->isCreative()){
+			return false;
```
dktapps: does this method have a return value? if so, you haven't returned anything below.

iPocket: returning void, ok?

PR #2229, opened 2016-11-21
Fix CraftItemEvent not return inputs

laurobeleche:
### Description
The getInput () function returned an empty array, a simple error to correct itself, just returning it to the correct variable.

### Reason to modify
Now the variable returns the requested correct value

### Tests & Reviews
Tested on my server, working properly.


laurobeleche:
And now?

PR #2231, opened 2016-11-23
Fix #2218

### LINKED ISSUE #2218: EntityArmorChangeEvent called in a wrong event.
#### Issue description

I've just found out that EntityArmorChangeEvent is being called during EntityDamageByBlockEvent which is kinda bad for people who want to get "Entity - while - changing - armor -> 's entity"

Try EntityArmorChangeEvent and see how its being called when the player is hit by cactus.

#### OS and versions

* Genisys: 55ae11e5acf93b753129343a3888cd4cf007da05
* PHP: 7.0.13
* Server OS: Ubuntu 14.04 64-bit (LTS)
* Game version: 0.16.1

```
Console is clean.
```

laurobeleche:
### Description
To fix de issue #2218

### Reason to modify
No longer calls the EntityArmorChangeEvent during the EntityDamageByBlockEvent

### Tests & Reviews
I tested it on my server, and it did not cause any kind of problem, working properly.

### DIFF
--- a/src/pocketmine/event/entity/EntityDamageEvent.php
+++ b/src/pocketmine/event/entity/EntityDamageEvent.php
@@ -304,9 +304,7 @@ class EntityDamageEvent extends EntityEvent implements Cancellable{
 				foreach($this->usedArmors as $index => $cost){
 					$i = $this->entity->getInventory()->getArmorItem($index);
 					if($i->isArmor()){
-						/** @var Armor $i */
-						$i->useOn($i, $cost);
-						$this->entity->getInventory()->setArmorItem($index, $i);
+						$this->entity->getInventory()->damageArmor($index, $cost);
 					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


GenisysFTW:
Good job mate. One thing,
Genisys generally doesn't accept pull requests having multiple commits. Try using got bash to combine your commits into one :)

laurobeleche:
Thanks, but I'm new to git and I do not know how to do this, can you explain to me that I do clean pulls

Muqsit:
Nice, but looks like Genisys will need more changes. Almost every entity damages call up EntityArmorChangeEvent...or atleast that's what I think, since EntityExplodeEvent does the same.

laurobeleche:
It will not be necessary since the location of the fix is in the EntityDamageEvent, these other classes are extended by the EntityDamageEvent

But thanks for watching

PR #2236, opened 2016-11-25
Small edit

ondraczsk:
### Description

Small edit, can be helpful for people

Not big but helpful


ondraczsk:
tell me why

ishitatsuyuki:
Do not touch the things that already work, you're just making it more confusing by adding redundant content.

dktapps:
In two words: totally unnecessary.

PR #2261, opened 2016-12-09
addEnchantment() returns the class.

IrishPacks:
### Description
This pull request is to simply have the item return the class after adding an enchant
instead of returning void.

### Tests & Reviews

 Works splendid.

### DIFF
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -588,8 +588,9 @@ class Item implements ItemIds, \JsonSerializable{
 				"lvl" => new ShortTag("lvl", $ench->getLevel())
 			]);
 		}
-
+		
 		$this->setNamedTag($tag);
+		return $this;
 	}
 
 	/**
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -588,7 +588,7 @@ class Item implements ItemIds, \JsonSerializable{
 				"lvl" => new ShortTag("lvl", $ench->getLevel())
 			]);
 		}
-		
+
 		$this->setNamedTag($tag);
 		return $this;
 	}


src/pocketmine/item/Item.php
```diff
@@ -588,8 +588,9 @@ public function addEnchantment(Enchantment $ench){
 				"lvl" => new ShortTag("lvl", $ench->getLevel())
 			]);
 		}
-
+
```
dktapps: Formatting

IrishPacks: Yikes. My bad. I'll fix

IrishPacks:
Simplicity and for API purposes so you can have an array of items all with the method addEnchantment at the end and it'll still return the class. I think it's a needed change imo.

ishitatsuyuki:
I don't like Java-style. It's okay in a garbage collected style language but not C-like languages. How does the other API acts? I think PM API doesn't return $this.

dktapps:
@ishitatsuyuki only for crap like crafting recipe registration, and that's an awful mess anyway.

A cleaner way to do it that doesn't involve `addEnchantment(blah)->addEnchantment(blah)->addEnchantment(blah)->addEnchantment(blah)->addEnchantment(blah)->addEnchantment(blah)->addEnchantment(blah)->addEnchantment(blah)->addEnchantment(blah)->addEnchantment(blah)->addEnchantment(blah)->addEnchantment(blah)`, which looks terrible in code, is to just use a simple foreach across an array of enchantments.

IrishPacks:
If  have an array of items such as $items = [Item::get(Item::BOOK)->addEnchantment(Enchantment::getEnchantment(Enchantment::PROTECTION)->setLevel(1), Item::get(Item::BOOK)->addEnchantment(Enchantment::getEnchantment(Enchantment::PROTECTION)->setLevel(2)];

foreach($items as $item){
//Assuming that $player is a player
$player->getInventory()->addItem($item);
}

It makes it easier because you shouldn't apply the same enchantment to different items. Tools need different enchantments, etc. So a foreach loop isn't the best way in my opinion.

PR #2326, opened 2017-01-21
Clean Player.php

Muqsit:
### Description
A cleanup of Player.php.

Removed deprecated exp functions.
Changed identity operator to equality wherever '==' was unneeded.
Used foreach(array_keys($array) as $k) instead of foreach($array as $k => $v) where $v was unneeded.

### Reason to modify
The usage of better PHP methods, eliminating useless $vars and more.

### Tests & Reviews

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -311,7 +311,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	}
 
 	public function setFishingHook(FishingHook $entity = null){
-		if($entity == null and $this->fishingHook instanceof FishingHook){
+		if($entity === null and $this->fishingHook instanceof FishingHook){
 			$this->fishingHook->close();
 		}
 		$this->fishingHook = $entity;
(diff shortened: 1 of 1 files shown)


src/pocketmine/Player.php
```diff
@@ -832,15 +731,15 @@ public function getInAirTicks(){
 	protected function switchLevel(Level $targetLevel){
 		$oldLevel = $this->level;
 		if(parent::switchLevel($targetLevel)){
-			foreach($this->usedChunks as $index => $d){
+			foreach(array_keys($this->usedChunks) as $index){
```
dktapps: I don't think you should change this. It will cause merge conflicts updating from PocketMine-MP, and function calls are expensive.

src/pocketmine/Player.php
```diff
@@ -940,7 +839,7 @@ protected function sendNextChunk(){
 		Timings::$playerChunkSendTimer->startTiming();
 
 		$count = 0;
-		foreach($this->loadQueue as $index => $distance){
+		foreach(array_keys($this->loadQueue) as $index){
```
dktapps: Same here

src/pocketmine/Player.php
```diff
@@ -1476,7 +1375,7 @@ public function setDataProperty($id, $type, $value){
 	}
 
 	protected function checkGroundState($movX, $movY, $movZ, $dx, $dy, $dz){
-		if(!$this->onGround or $movY != 0){
+		if(!$this->onGround or $movY !== 0){
```
dktapps: Bad idea. These vars are usually floats. 0.0 !== 0, but 0.0 == 0

src/pocketmine/Player.php
```diff
@@ -1624,7 +1523,7 @@ protected function processMovement($tickDiff){
 			}
 		}
 
-		if(!$revert and $distanceSquared != 0){
+		if(!$revert and $distanceSquared !== 0){
```
dktapps: same here

src/pocketmine/Player.php
```diff
@@ -1708,7 +1607,7 @@ protected function processMovement($tickDiff){
 			}
 
 			$this->speed = $from->subtract($to);
-		}elseif($distanceSquared == 0){
+		}elseif($distanceSquared === 0){
```
dktapps: and here

src/pocketmine/Player.php
```diff
@@ -2123,12 +2022,12 @@ protected function processLogin(){
 		$pk->lightningLevel = 0;
 		$pk->commandsEnabled = 1;
 		$pk->unknown = "UNKNOWN";
-		$pk->worldName = $this->server->getMotd();
+		$pk->worldName = $this->server->getMotd();//$this->level->getName()
```
dktapps: any reason to add this? It cannot be changed at runtime.

src/pocketmine/Player.php
```diff
@@ -4253,14 +4154,13 @@ public function isLoaderActive(){
...
-	}
+    /**
+     * @param Effect $effect
+     */
+    public function addEffect(Effect $effect){//Overwrite
+        if($effect->isBad() && $this->isCreative()){
+            return;
+        }
+        parent::addEffect($effect);
+    }
```
dktapps: why space indentation?

src/pocketmine/Player.php
Muqsit: Oh!

Muqsit:
Will close it for now..

PR #2364, opened 2017-02-21
Fix underwater breath

### LINKED ISSUE #2343: Breath underwater
#### Issue description

Breath underwater is decreasing way too fast

Just go underwater
#### OS and versions

* Genisys:Genisys-1ed05b7
* PHP:7.0.14
* Server OS:Win7
* Game version: 1.0.0.16

```
```

diamond-gold:
### Description

To fix #2343 

### Reason to modify

Fix #2343, player's breath is decreasing way too fast

### Tests & Reviews

Tested the code and it works correctly with enchantment and without enchantment

### DIFF
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -182,7 +182,8 @@ abstract class Living extends Entity implements Damageable{
 				$ev = new EntityDamageEvent($this, EntityDamageEvent::CAUSE_SUFFOCATION, 1);
 				$this->attack($ev->getFinalDamage(), $ev);
 			}
-
+			$maxAir = 400 + $EnchantL * 300;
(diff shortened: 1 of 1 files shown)


dktapps:
👍

Muqsit:
L198:
You declare a variable to be an alias of
400 + $EnchantL * 300

L200:
You don't use the variable at all. You rather type in 400 + $EnchantL * 300. The variable declaration was hence useless.

Why does it have to be 400 + $EnchantL * 300, by the way?

Muqsit:
L199: Code formatting..
I think there's a much better way to write this. If you overview it, you'll find out how much of code have you repeated. Avoid repetition to its best.

diamond-gold:
400 + $EnchantL * 300
Its by default 20 seconds, 400 ticks, and increases 15 seconds(300 ticks) per level

PR #2373, opened 2017-03-02
fix genisys api

caspervanneck:
### Description

for genisys api dont work

### Reason to modify

i try to fix it. i test it and it works great

### Tests & Reviews

yes i have test  it work and it breaks nothing


JackNoordhuis:
The API checks exist for a reason.

HAN-105:
what API (from pocketmine) is using ? for devtools ? the latest version .phar from gitlab

PR #2379, opened 2017-03-06
Add Ender pearls

JackNoordhuis:
### Description
Add the basic vanilla Ender Pearl implementation

### Reason to modify
Ender Pearls are a vanilla feature commonly used by servers.

### Tests & Reviews
Thoroughly tested by myself, works exactly like vanilla.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2529,7 +2529,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 							$thrownPotion->spawnToAll();
 						}
 					}elseif($item->getId() === Item::ENDER_PEARL){
-						if(($time = microtime(true)) - $this->lastEnderPearlUse >= 1){
+						if(floor(($time = microtime(true)) - $this->lastEnderPearlUse) >= 1){
 							$nbt = new CompoundTag("", [
 								"Pos" => new ListTag("Pos", [
 									new DoubleTag("", $this->x),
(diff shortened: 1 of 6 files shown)


Pizzahaet:
I love you for this (no homo)

src/pocketmine/Player.php
```diff
@@ -2524,6 +2528,45 @@ public function handleDataPacket(DataPacket $packet){
 						}else{
 							$thrownPotion->spawnToAll();
 						}
+					}elseif($item->getId() === Item::ENDER_PEARL){
+						if(floor(($time = microtime(true)) - $this->lastEnderPearlUse) >= 1){
```
ishitatsuyuki: Floor seems redundant to me.

diamond-gold:
Bug: if ProjectileLaunchEvent is cancelled for ender pearl, it damages the player and teleports the player to the same position

JackNoordhuis:
The projectile bug seems to happen with thrown potions as well, this should be fixed in a separate commit as this pull request is simply to add ender pearls, not fix an existing projectile bug. @diamond-gold 

@ishitatsuyuki I can remove the floor if you like, I add it to all my micro time comparisons to make them more accurate.

ishitatsuyuki:
How can it mathematically make it more accurate ..?

JackNoordhuis:
In some sense it's actually less accurate, but we only want the check to pass if it's been more than x seconds so anything less doesn't matter. Rounding the value to the lowest whole number was faster than comparing float's in PHP5 but I just did some benchmark tests that have shown this to be the opposite with PHP7.

diamond-gold:
@JackNoordhuis The bug is because when you cancel ProjectileLaunchEvent, you called $enderpearl->kill()
And inside Enderpearl::kill() you damage the player and teleport the player, it is not an existing bug

JackNoordhuis:
@diamond-gold I'm aware of what's causing the issue, I've been busy cleaning up the projectile code as a whole. A fix will be applied soon.

PR #2381, opened 2017-03-07
Timings v1 Alternative

AppleDevelops:
### Description

It uses a fork of timings v1.

### Reason to modify

Timings v1 no longer works, and people seem to dislike timings v2.
### Tests & Reviews


Muqsit:
IMO, Genisys (the person having access to the official Genisys website) should host Timings themselves; because you really can't count on using a third-party, unless the third party was you yourself.

dktapps:
So the suggestion is, we hardcode an alternative. And when this server goes down too, we have to hardcode another one? I wonder, why does that sound like a bad idea... 🤔

PR #2390, opened 2017-03-16
Re-implement ItemFrameDropItemEvent, fix #2389

### LINKED ISSUE #2389: Players can steal items in "Item Frame"
The players has **no op** in my server can steal items in the **Item Frame**.
 - 16/03/2017

No code..

Muqsit:
### Description

It was previously possible to control the behavior of item frame droppings through PlayerInteractEvent (<= v0.16). This PR fixes ItemFrameDropItemEvent not being triggered.

### Reason to modify

This will help developers extend the API to modify interactions with Item Frames. Cancellation of the event now prevents the item frame from dropping the item.

### Tests & Reviews

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -42,6 +42,7 @@ use pocketmine\entity\Minecart;
 use pocketmine\entity\Projectile;
 use pocketmine\entity\ThrownExpBottle;
 use pocketmine\entity\ThrownPotion;
+use pocketmine\event\block\ItemFrameDropItemEvent;
 use pocketmine\event\entity\EntityCombustByEntityEvent;
 use pocketmine\event\entity\EntityDamageByBlockEvent;
 use pocketmine\event\entity\EntityDamageByEntityEvent;
@@ -3321,7 +3322,9 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 				$tile = $this->level->getTile($this->temporalVector->setComponents($packet->x, $packet->y, $packet->z));
 				if($tile instanceof ItemFrame){
-					if($this->isSpectator()){
+					$this->server->getPluginManager()->callEvent($ev = new ItemFrameDropItemEvent($this, $tile->getBlock(), $tile, $tile->getItem()));
+
+					if($this->isSpectator() or $ev->isCancelled()){
 						$tile->spawnTo($this);
 						break;
 					}
--- a/src/pocketmine/event/block/ItemFrameDropItemEvent.php
+++ b/src/pocketmine/event/block/ItemFrameDropItemEvent.php
@@ -34,7 +34,7 @@ class ItemFrameDropItemEvent extends BlockEvent implements Cancellable{
 	private $player;
 	/** @var  Item */
 	private $item;
-	/** @var  ItemFrame */
+	/** @var ItemFrame $itemFrame */
 	private $itemFrame;
 
 	public function __construct(Player $player, Block $block, ItemFrame $itemFrame, Item $item){


dktapps:
Do note however that this will be an API fluctuation, since that will change to generic interaction + item drop in the future.

src/pocketmine/event/block/ItemFrameDropItemEvent.php
```diff
@@ -34,7 +34,7 @@ class ItemFrameDropItemEvent extends BlockEvent implements Cancellable{
 	private $player;
 	/** @var  Item */
 	private $item;
-	/** @var  ItemFrame */
+	/** @var ItemFrame $itemFrame */
```
dktapps: This addition is not necessary.

ishitatsuyuki:
Do not spam me with review request. I will review some when I'm extremely bored. I don't regularly work on Genisys.

src/pocketmine/Player.php
```diff
@@ -3321,7 +3322,9 @@ public function handleDataPacket(DataPacket $packet){
 
 				$tile = $this->level->getTile($this->temporalVector->setComponents($packet->x, $packet->y, $packet->z));
 				if($tile instanceof ItemFrame){
-					if($this->isSpectator()){
+					$this->server->getPluginManager()->callEvent($ev = new ItemFrameDropItemEvent($this, $tile->getBlock(), $tile, $tile->getItem()));
+
+					if($this->isSpectator() or $ev->isCancelled()){
```
dktapps: Maybe better to just setCancelled() if the player is spectator instead of hardcoding this. Then plugins can override it.

src/pocketmine/event/block/ItemFrameDropItemEvent.php
Muqsit: Descriptive comments..Just OCD.

src/pocketmine/Player.php
Muqsit: That will make it possible to allow spectators to modify terrain. That shouldn't be possible.

dktapps: not in vanilla... but be imaginative. 😉 Plugins should be able to alter this behaviour if they choose to. I dislike the inflexibility of the hardcoded gamemodes.

Muqsit: You make a point. 😐

JackNoordhuis:
Wouldn't it be better to have PlayerInteractEvent be called instead of adding a new event? Just seems sort of useless adding more and more events just because of some additional functionality.

Vaivez66:
I agree with @JackNoordhuis

dktapps:
@JackNoordhuis that would be a future-proof method, yes. This event and packet will be removed in the future.

dktapps:
@JackNoordhuis I don't believe there is any need for a separate event at all. You can simply use PlayerInteractEvent with the left-click action and check $ev->getBlock() instanceof ItemFrameBlock if you're really bothered.

dktapps:
The only thing I consider out of the ordinary for item frame drops is that they can have varying chances of dropping items, and this event doesn't even cater for that. It's unnecessary overcomplication.

JackNoordhuis:
@dktapps Having the seperate event allows you to add the extra data, like the item entity.

dktapps:
@JackNoordhuis but it doesn't... have you actually looked at the implementation? There's nothing you can do with that event that you can't do with PlayerInteractEvent, provided it was correctly called.

Muqsit:
IMO, there should be a whole new PlayerInteractTileEvent.

dktapps:
@Muqsit Tiles are simply blocks with special properties. They are more correctly referred to as block entities. I don't think there is a need for that.

PR #2400, opened 2017-04-05
Available forks/spoons of genisys information

poustmal:
Link for a fork of genisys and a bit information for players that they don't know that there is another software a bit like this.


dktapps:
Don't. Be. An. Idiot.

